PR #4, opened 2019-11-25
Add call for the FileChooser::SaveFiles portal method

mjog:
Adds xdp_portal_save_files and xdp_portal_save_files_finish functions.

### DIFF
--- a/libportal/filechooser.c
+++ b/libportal/filechooser.c
@@ -40,7 +40,7 @@ typedef struct {
   XdpPortal *portal;
   XdpParent *parent;
   char *parent_handle;
-  gboolean save_mode;
+  char *method;
   char *title;
   gboolean modal;
   gboolean multiple;
@@ -77,6 +77,7 @@ file_call_free (FileCall *call)
   g_object_unref (call->portal);
   g_object_unref (call->task);
 
+  g_free (call->method);
   g_free (call->title);
   g_free (call->current_name);
   g_free (call->current_folder);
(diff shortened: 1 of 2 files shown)


mjog:
Along with https://github.com/flatpak/xdg-desktop-portal/pull/368 and https://github.com/flatpak/xdg-desktop-portal-gtk/pull/236, this is ready to commit barring any review comments.

libportal/filechooser.c
```diff
@@ -40,13 +40,14 @@ typedef struct {
   XdpPortal *portal;
   XdpParent *parent;
   char *parent_handle;
-  gboolean save_mode;
+  char *mode;
```
matthiasclasen: I think this should be called method

libportal/filechooser.c
```diff
@@ -437,3 +443,100 @@ xdp_portal_save_file_finish (XdpPortal *portal,
...
   return ret ? g_variant_ref (ret) : NULL;
 }
+
+/**
+ * xdp_portal_save_files:
+ * @portal: a #XdpPortal
+ * @parent: (nullable): parent window information
+ * @title: title for the file chooser dialog
+ * @modal: whether the dialog should be modal
+ * @current_name: (nullable): suggested filename
```
matthiasclasen: The implementation doesn't use current_name, so we should drop it from the api here

libportal/filechooser.c
mjog: Done, cheers.

libportal/filechooser.c
mjog: Hmm, I left this in the actual portal call since I figured applications might want to suggest a folder name, but I guess that probably doesn't make sense at least for GTK? I guess on that basis I should take it out of the portal impls as well?

matthiasclasen: I don't think the gtk file chooser lets you suggest folder names, or does it ? If it does, sure, keep it in. But then the docs should be clearer that this is the suggested name for a newly created directory containing the files to save.

mjog:
> This `g_free` looks bad; method is only assigned from static strings.

You're right - thanks. I left it out initially but when I came back to it I noticed it was missing and "fixed" it. Just posted https://github.com/flatpak/libportal/pull/5 to fix.

Issue #8, opened 2019-12-16
correct way to include headers is ambiguous


smcv:
Are users of libportal meant to `#include <libportal/portal.h>` or `#include <portal.h>`?

If the former, then `@includedir@` in the pkg-config file should probably be set to `"${prefix}/" + get_option("includedir")`, i.e. normally `/usr/include`, so that if the user has installed libportal in (for example) `/opt/flatpak-stuff`, they will get the correct `-I/opt/flatpak-stuff/include` to be able to use `#include <libportal/portal.h>` without error. It would probably also be a good idea for cross-inclusions between headers to use the format `#include "account.h"` or `#include <libportal/account.h>`, to avoid accidentally including a header like `/usr/include/account.h` if some other package includes that header.

Or, if the latter, then the headers should probably be renamed to something less namespace-polluting, for example `portal-account.h` instead of `account.h`.

matthiasclasen:
Pretty sure I was going for <libportal/portal.h>

smcv:
Thanks, I'll try to put together a PR if you don't get there first.

matthiasclasen:
I'll do it. Need to do some work here anyway

PR #9, opened 2019-12-17
Clean up includes

matthiasclasen:
The intention was to have application code use
changes to make that actually work.

Fixes: https://github.com/flatpak/libportal/issues/8


smcv:
Looks good to me!

PR #14, opened 2019-12-20
Convert some debug printfs into g_debug()

smcv:
g_print() is problematic for library users that have machine-readable
output like TAP or JSON on stdout, whereas the g_log() family are
easy to divert to stderr in such cases.


hadess:
Looks good to me.

PR #18, opened 2020-01-14
Add Qt support

grulja:
One thing I'm not sure with is whether in case of wayland, we should at least return some handle, otherwise all applications will fail to use libportal.

### DIFF
--- a/libportal/portal-qt5.h
+++ b/libportal/portal-qt5.h
@@ -36,11 +36,17 @@ static inline gboolean _xdp_parent_export_qt (XdpParent *parent,
         return TRUE;
       }
     }
-
-  /* TODO: QtWayland doesn't support xdg-foreign protocol yet
-   * Upstream bugs: https://bugreports.qt.io/browse/QTBUG-73801
-   *                https://bugreports.qt.io/browse/QTBUG-76983
-   */
+  else
+    {
+      /* TODO: QtWayland doesn't support xdg-foreign protocol yet
+       * Upstream bugs: https://bugreports.qt.io/browse/QTBUG-73801
+       *                https://bugreports.qt.io/browse/QTBUG-76983
+       */
+      g_warning ("QtWayland doesn't support xdg-foreign protocol yet");
+      g_autofree char *handle = g_strdup ("");
+      callback (parent, handle, data);
+      return TRUE;
+    }
 
   g_warning ("Couldn't export handle, unsupported windowing system");
   return FALSE;
--- a/libportal/portal-qt5.h
+++ b/libportal/portal-qt5.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2020, Jan Grulich
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


libportal/utils.c
```diff
@@ -29,6 +29,7 @@ _xdp_parent_copy (XdpParent *source)
   parent->parent_export = source->parent_export;
   parent->parent_unexport = source->parent_unexport;
   g_set_object (&parent->object, source->object);
+  parent->data = source->data;
```
hadess, line 32: A better commit message would be appreciated.

libportal/portal-qt5.h
```diff
@@ -36,11 +36,17 @@ static inline gboolean _xdp_parent_export_qt (XdpParent *parent,
...
-
-  /* TODO: QtWayland doesn't support xdg-foreign protocol yet
-   * Upstream bugs: https://bugreports.qt.io/browse/QTBUG-73801
-   *                https://bugreports.qt.io/browse/QTBUG-76983
-   */
+  else
+    {
+      /* TODO: QtWayland doesn't support xdg-foreign protocol yet
+       * Upstream bugs: https://bugreports.qt.io/browse/QTBUG-73801
+       *                https://bugreports.qt.io/browse/QTBUG-76983
```
hadess, line 43: Any movement on those?

hadess:
Can you rebase your PR properly so that merge branches don't appear in it? Also `Do not forgot to install portal-qt5 header` should probably be merged into `Add Qt support`.

Looks fine at a glance, though this is all probably blocking on the Qt Wayland support for parenting the dialogues. Did you have plans to also write a test application that can be sandboxed, similar to the GTK one?

libportal/portal-qt5.h
grulja: Unfortunately no and most likely this won't change in near future as the main QtWayland developer left Qt Company and QtWayland is now without a maintainer.

grulja:
> Can you rebase your PR properly so that merge branches don't appear in it? Also `Do not forgot to install portal-qt5 header` should probably be merged into `Add Qt support`.
> 

Done.

> Looks fine at a glance, though this is all probably blocking on the Qt Wayland support for parenting the dialogues. Did you have plans to also write a test application that can be sandboxed, similar to the GTK one?

Not sure if we should be blocking on that, at least X11 support works just fine. I will submit a better test application with my Qt wrapper for libportal, where I test every single portal to test the wrapper.

hadess:
> Not sure if we should be blocking on that, at least X11 support works just fine.

But it fails on Wayland every time, right? I would expect things to at least somewhat work, but I'll defer to @matthiasclasen on that.

grulja:
> > Not sure if we should be blocking on that, at least X11 support works just fine.
> 
> But it fails on Wayland every time, right? I would expect things to at least somewhat work, but I'll defer to @matthiasclasen on that.

It doesn't fail on Wayland, we already discussed that with @matthiasclasen, that's why I set an empty handle, to avoid the failure.

hadess:
> It doesn't fail on Wayland, we already discussed that with @matthiasclasen, that's why I set an empty handle, to avoid the failure.

OK, in which case we'd probably just want to downgrade the `g_warning()` to a `g_warning_once()`, but that would bump the dependency to glib 2.64. @matthiasclasen?

libportal/portal-qt5.h
plfiorini: This is a new feature, it's too late for Qt 5.15. Since it's the last 5.x release, you have to wait at least for Qt 6.0.
If somebody can make a patch there is people that will review and approve it even though Johan left.

grulja:
Ping?

matthiasclasen:
I would use an #ifdef to avoid bumping the glib dep just for g_warning_once

PR #19, opened 2020-01-16
updates: fix stray debug g_print()

chergert:


smcv:
See also #14.

hadess:
Hope you don't mind, but I merged https://github.com/flatpak/libportal/pull/14 instead.

PR #20, opened 2020-02-07
Build fixes

jtojnar:
On Nix, the library failed to build because `portal-enums.h` was installed into an incorrect directory. Additionally, the `.pc` file had incorrectly joined paths.

Also, checking the headers, I do not see `gio-unix-2.0` being used, only `gio-2.0` so I left the former in `Requires.private`. This is the diff of the pc file:

```diff
--- /nix/store/wvzydk8lzvf4klsfc0145pi9nyzv48fh-libportal-0.3-dev/lib/pkgconfig/libportal.pc
+++ /nix/store/vpx6328ks5dgwzrf0wpcn3scn6rx4w65-libportal-0.3-dev/lib/pkgconfig/libportal.pc
@@ -1,10 +1,11 @@
-prefix=/nix/store/lkl9cjhgxdxzi8nr4njywp65xmycpgd0-libportal-0.3
+prefix=/nix/store/r29z11ah41ny9zmzs1w5xji6zdl6lp4q-libportal-0.3
 libdir=${prefix}/lib
-includedir=/nix/store/wvzydk8lzvf4klsfc0145pi9nyzv48fh-libportal-0.3-dev//nix/store/wvzydk8lzvf4klsfc0145pi9nyzv48fh-libportal-0.3-dev/foo
+includedir=/nix/store/vpx6328ks5dgwzrf0wpcn3scn6rx4w65-libportal-0.3-dev/foo
 
 Name: libportal
 Description: Portal API wrappers
 Version: 0.3
-Requires: gio-unix-2.0
+Requires: gio-2.0
+Requires.private: gio-unix-2.0
 Libs: -L${libdir} -lportal
 Cflags: -I${includedir}

```


libportal/meson.build
```diff
@@ -63,6 +63,12 @@ libportal = library('portal',
                     install: true,
                     dependencies: [gio_dep, gio_unix_dep])
 
+pkgconfig.generate(libportal,
+  description: 'Portal API wrappers',
+  name: 'libportal',
+  requires: [ gio_dep ],
```
hadess: That should be `gio_unix_dep` otherwise you're changing the `Requires` from `gio-unix-2.0` to `gio-2.0`. I don't know whether that's correct, but if it is, it needs to happen in a separate commit.

jtojnar: Should I fix it in the old `.pc.in` template explicitly? It is not really possible to re-create the previous incorrect file using the Meson’s `pkgconfig` module since it correctly adds `library`’s `dependencies` to `Requires.private` (https://mesonbuild.com/Pkgconfig-module.html#implicit-dependencies).

hadess: > Should I fix it in the old .pc.in template explicitly?

That would be best, thanks.

jtojnar: Done.

Issue #21, opened 2020-02-07
Wrong types used in SaveFile() method from FileChooser portal


grulja:
In **xdp_portal_save_file()** users are expected to pass **current_folder** and **current_file** as strings and they seem to be passed to the xdg-desktop-portal that way as well, but the portal expects them as byte array. You can see in documentation [1] that **"ay"** is expected. This results into failure when doing this call. 

[1] - https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.portal.FileChooser.xml#L213

grulja:
Looks **xdp_portal_save_files()** is the same case with **current_folder** parameter.

grulja:
Fixed with https://github.com/flatpak/libportal/pull/23.

PR #23, opened 2020-02-11
FileChooser: current_file and current_folder should be send as byte strings

grulja:


grulja:
This fixes https://github.com/flatpak/libportal/issues/21.

PR #24, opened 2020-03-18
Build fixes

hadess:


hadess:
The last commit didn't work in the original version, forgot the meson arguments.

hadess:
And I've removed `flatpak: Allow running build script from another dir` because running it from the same dir didn't work anymore.

PR #28, opened 2020-03-31
Add Makefile

Arxcis:
## Why?
I just started playing around with this project, which resulted in a `Makefile` for listing common operations.
If the project had a `Makefile` already, it would have been easier for me to get started 🚀
libportal-test


matthiasclasen:
Well, we're not using make, but meson.

I would accept a pr that adds build instructions to README.md

PR #30, opened 2020-05-20
camera: Support selecting a camera

matthiasclasen:
Add a variant of xdp_portal_access_camera_finish called
xdp_portal_access_camera_finish_streams that returns
information about what camera was selected, in the same
format that screencast sessions return stream information.

### DIFF
--- a/libportal/camera.c
+++ b/libportal/camera.c
@@ -122,7 +122,15 @@ response_received (GDBusConnection *bus,
   g_variant_get (parameters, "(u@a{sv})", &response, &ret);
 
   if (response == 0)
-    g_task_return_boolean (call->task, TRUE);
+    {
+      GVariant *streams;
+
+      g_print ("ret: %s\n", g_variant_print (ret, FALSE));
+      g_variant_lookup (ret, "streams", "@a(ua{sv})", &streams);
+      g_object_set_data_full (G_OBJECT (call->task), "streams",
+                              g_variant_ref (streams), (GDestroyNotify)g_variant_unref);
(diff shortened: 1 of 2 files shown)


libportal/camera.c
```diff
@@ -276,6 +284,27 @@ xdp_portal_access_camera_finish (XdpPortal     *portal,
...
+GVariant *
+xdp_portal_access_camera_finish_streams (XdpPortal     *portal,
+                                         GAsyncResult  *result,
+                                         GError       **error)
+{
+  g_return_val_if_fail (XDP_IS_PORTAL (portal), FALSE);
+  g_return_val_if_fail (g_task_is_valid (result, portal), FALSE);
+  g_return_val_if_fail (g_task_get_source_tag (G_TASK (result)) == xdp_portal_access_camera, FALSE);
+
+  if (g_task_propagate_boolean (G_TASK (result), error))
```
GeorgesStavracas, line 296: Would it be necessary to retrieve `streams` before calling this? I think this could potentially drop the last reference to this GTask.

libportal/camera.c
```diff
@@ -122,7 +122,15 @@ response_received (GDBusConnection *bus,
   g_variant_get (parameters, "(u@a{sv})", &response, &ret);
 
   if (response == 0)
-    g_task_return_boolean (call->task, TRUE);
+    {
+      GVariant *streams;
+
+      g_print ("ret: %s\n", g_variant_print (ret, FALSE));
```
GeorgesStavracas, line 128: `g_debug`?

PR #40, opened 2021-06-17
CI: Add mono apt repository

GeorgesStavracas:
Copy-pasting [1]. This fixes the pre-build step.

[1] https://github.com/flatpak/flatpak-builder/commit/1898b36d1fae6f011e5465982154e18320549470

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo add-apt-repository ppa:alexlarsson/glib260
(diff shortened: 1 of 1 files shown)


smcv (approved):
I hate that it's apparently necessary to cargo-cult this around, but the base package set on Github's CI workers is not under our control, so this seems like a necessary evil.

(Not really a maintainer - or maybe I am? I am technically able to merge things in this repo as a result of being added to the Flatpak organization, but I don't want to step on the real maintainers' toes if they consider libportal's maintainer list to be smaller than Flatpak's.)

PR #41, opened 2021-06-18
[API Break] Add cursor mode to remote APIs

GeorgesStavracas:
There currently is no way to pass a cursor mode when creating
a screencast / remote desktop session, which is pretty limiting.

Add XdpCursorMode, and the corresponding parameters to the session
functions.

### DIFF
--- a/libportal/meson.build
+++ b/libportal/meson.build
@@ -57,8 +57,7 @@ install_headers(headers, subdir: 'libportal')
 
 libportal = library('portal',
                     src,
-                    version: '0.0.1',
-                    soversion: 0,
+                    version: '1.0.0',
                     include_directories: [top_inc, libportal_inc],
                     c_args: visibility_args,
                     install: true,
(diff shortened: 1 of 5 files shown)


TingPing:
Should bump soname for api break. Looks reasonable.

matthiasclasen:
Looks fine to me too. Will you bump the soname?

libportal/meson.build
```diff
@@ -58,7 +58,7 @@ install_headers(headers, subdir: 'libportal')
 libportal = library('portal',
                     src,
                     version: '0.0.1',
```
TingPing, line 60: This should be `1.0.0`.

You can actually remove `soversion` entirely as the first number of `version` is used.

GeorgesStavracas: Errr, apologies for the massive brainfart with the library version. Should be good to go now.

GeorgesStavracas:
Hm, maybe we should start appending the major version to the library name? E.g. `libportal-0`, `libportal-1`, etc.

TingPing:
I think we should do that by time we consider this API stable. I don't think we need to immediately do it.

TingPing (approved):

PR #45, opened 2021-06-28
Fix location updated signal

grulja:
1) All numbers should be type of double and not int
2) Connect to correct DBus object
3) Add missing description property


GeorgesStavracas (changes requested):
Can you please split these 3 changes in 3 commits, one for each change?

GeorgesStavracas (approved):

PR #46, opened 2021-06-30
Add libportal-qt

grulja:
Libportal-qt is a Qt wrapper for libportal. It uses very similar API, it's just more qute (Qt friendly).

### Qt wrappers
- [x] Account portal
- [x] Background portal
- [x] Camera portal
- [x] Email portal
- [x] FileChooser portal
- [x] Inhibit portal
- [x] Location portal
- [x] Notification portal
- [x] OpenUri portal
- [x] Print portal

### How does the API look like?
GLib
```
static void
opened_uri (GObject *object,
            GAsyncResult *result,
            gpointer data)
{
  /* Implementation of opened_uri response */
}

XdpOpenUriFlags flags = XDP_OPEN_URI_FLAG_NONE;
XdpParent *parent = xdp_parent_new_gtk (GTK_WINDOW (win));
xdp_portal_open_uri (win->portal, parent, "https://github.com/flatpak/libportal", flags, NULL, opened_uri, win);
```
Qt
```
Xdp::Parent xdpParent(windowHandle());
Xdp::openUri(xdpParent, QStringLiteral("https://github.com/flatpak/libportal"), Xdp::OpenUriFlag::None);
connect(Xdp::notifier(), &Xdp::Notifier::openUriResponse, this, [=] (const Xdp::Response &response) {
     if (response.isSuccess()) { // DO SOMETHING }
});
```

### Other benefits
This Qt wrapper also makes it more convenient to use, because the GLib implementation takes some arguments as GVariants and forces everyone to look into the documentation for all possible values they can pass there, while I wrote convenient classes with getters/setters for each option.

E.g. in the FileChooser portal:
**GLib**: `xdg_portal_open_file()` takes filters as `GVariant`
**Qt**: `Xdp::openFile()` takes `FileChooserFilter` class as an argument with getters/setters to set `id`, `label` and `rule`.

### DIFF
--- a/libportal-qt/email.cpp
+++ b/libportal-qt/email.cpp
@@ -1,0 +1,116 @@
+/*
+ * Copyright (C) 2020-2021, Jan Grulich
+ *
+ * This file is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, version 3.0 of the
+ * License.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * SPDX-License-Identifier: LGPL-3.0-only
+ */
+
+#include "email.h"
+#include "parent.h"
+#include "parent_p.h"
+#include "portal_p.h"
+
+#include <string.h>
+#include <iostream>
+
+namespace Xdp
+{
+
+void PortalPrivate::composeEmail(const Parent &parent, const QStringList &addresses, const QStringList &cc,
+                                 const QStringList &bcc, const QString &subject, const QString &body,
+                                 const QStringList &attachments, EmailFlags flags)
+{
+    int i = 0;
+    char *addressesArray[addresses.size() + 1];
+    char *ccArray[cc.size() + 1];
+    char *bccArray[bcc.size() + 1];
+    char *attachmentsArray[attachments.size() + 1];
+
+    while (i < addresses.size()) {
+        addressesArray[i] = new char[addresses.at(i).length() + 1];
+        strcpy(addressesArray[i], addresses.at(i).toLatin1());
+        addressesArray[i][addresses.at(i).length()] = '\0';
+        i++;
+    }
+    addressesArray[i] = nullptr;
+
+    i = 0;
+    while (i < cc.size()) {
+        ccArray[i] = new char[cc.at(i).length() + 1];
+        strcpy(ccArray[i], cc.at(i).toLatin1());
+        ccArray[i][cc.at(i).length()] = '\0';
+        i++;
+    }
+    ccArray[i] = nullptr;
+
+    i = 0;
+    while (i < bcc.size()) {
+        bccArray[i] = new char[bcc.at(i).length() + 1];
+        strcpy(bccArray[i], bcc.at(i).toLatin1());
+        bccArray[i][bcc.at(i).length()] = '\0';
+        i++;
+    }
+    bccArray[i] = nullptr;
+
+    i = 0;
+    while (i < attachments.size()) {
+        attachmentsArray[i] = new char[attachments.at(i).length() + 1];
+        strcpy(attachmentsArray[i], attachments.at(i).toLatin1());
+        attachmentsArray[i][attachments.at(i).length()] = '\0';
+        i++;
+    }
+    attachmentsArray[i] = nullptr;
+
+    xdp_portal_compose_email(m_xdpPortal, parent.d_ptr->m_xdpParent, addressesArray, ccArray, bccArray, const_cast<char*>(subject.toUtf8().constData()),
+                             const_cast<char*>(body.toUtf8().constData()), attachmentsArray, static_cast<XdpEmailFlags>((int)flags), nullptr, composedEmail, this);
+
+    i = 0;
+    while (i < addresses.size()) {
+        delete addressesArray[i];
+        i++;
+    }
+
+    i = 0;
+    while (i < cc.size()) {
+        delete ccArray[i];
+        i++;
+    }
+
+    i = 0;
+    while (i < bcc.size()) {
+        delete bccArray[i];
+        i++;
+    }
+
+    i = 0;
+    while (i < attachments.size()) {
+        delete attachmentsArray[i];
+        i++;
+    }
+}
+
+void PortalPrivate::composedEmail(GObject *object, GAsyncResult *result, gpointer data)
+{
+    XdpPortal *xdpPortal = XDP_PORTAL(object);
+    PortalPrivate *portalPrivate = static_cast<PortalPrivate*>(data);
+    g_autoptr(GError) error = nullptr;
+
+    bool ret = xdp_portal_compose_email_finish(xdpPortal, result, &error);
+
+    Response response(ret, error);
+    portalPrivate->composeEmailResponse(response);
+}
+
+}
--- a/libportal-qt/filechooser.h
+++ b/libportal-qt/filechooser.h
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2020-2021, Jan Grulich
+ *
+ * This file is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, version 3.0 of the
+ * License.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * SPDX-License-Identifier: LGPL-3.0-only
+ */
+
+#ifndef LIBPORTALQT_FILECHOOSER_H
+#define LIBPORTALQT_FILECHOOSER_H
+
+#include "libportalqt_export.h"
+
+#include <QFlags>
+#include <QMap>
+#include <QStringList>
+#include <QSharedPointer>
+
+namespace Xdp
+{
+
+enum class OpenFileFlag
+{
+    None = 0x0,
+    Multiple = 0x1
+};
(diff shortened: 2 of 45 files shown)


grulja:
My plan is to add at least support for `notification portal` and `print portal` as the rest is not that widely used and I don't want to block this PR on them.

May I ask you to start with review of the current portal implementations while I keep finishing the remaining ones?

CC: @matthiasclasen @GeorgesStavracas @hadess @felipeborges

hadess:
> May I ask you to start with review of the current portal implementations while I keep finishing the remaining ones?

It might be a good idea to get someone familiar with Qt/KDE to review this, it's mostly client code.

grulja:
Maybe @aleixpol can help with this?

libportal-qt/email.cpp
```diff
@@ -0,0 +1,116 @@
...
+    i = 0;
+    while (i < attachments.size()) {
+        attachmentsArray[i] = new char[attachments.at(i).length() + 1];
+        strcpy(attachmentsArray[i], attachments.at(i).toLatin1());
+        attachmentsArray[i][attachments.at(i).length()] = '\0';
+        i++;
+    }
+    attachmentsArray[i] = nullptr;
+
+    xdp_portal_compose_email(m_xdpPortal, parent.d_ptr->m_xdpParent, addressesArray, ccArray, bccArray, const_cast<char*>(subject.toStdString().c_str()),
```
aleixpol, line 76: Rather than toStdString, do .toUtf8().constData()

libportal-qt/email.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+void PortalPrivate::composedEmail(GObject *object, GAsyncResult *result, gpointer data)
+{
+    XdpPortal *xdpPortal = XDP_PORTAL(object);
+    PortalPrivate *portalPrivate = static_cast<PortalPrivate*>(data);
+    g_autoptr(GError) error = nullptr;
+
+    bool ret = xdp_portal_compose_email_finish(xdpPortal, result, &error);
+
+    Response response(ret, error ? QString(error->message) : QString());
```
aleixpol, line 112: This might warn on some systems. Use `QString::fromUtf8()`

libportal-qt/filechooser.h
```diff
@@ -0,0 +1,140 @@
...
+    bool isValid() const;
+
+    QString id() const;
+    void setId(const QString &id);
+
+    QString label() const;
+    void setLabel(const QString &label);
+
+    QMap<QString, QString> options() const;
+    void addOption(const QString &id, const QString &label);
```
aleixpol, line 124: Are these options really any string? Would it make sense to make them an enum or actual properties? or the idea is that they will get extended over time?

libportal-qt/inhibit.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    XdpPortal *xdpPortal = XDP_PORTAL(object);
+    PortalPrivate *portalPrivate = static_cast<PortalPrivate*>(data);
+    g_autoptr(GError) error = nullptr;
+
+    int ret = xdp_portal_session_inhibit_finish(xdpPortal, result, &error);
+
+    QVariantMap responseData;
+    responseData.insert(QStringLiteral("id"), ret);
+    Response response(ret, error ? QString(error->message) : QString(), responseData);
```
aleixpol, line 45: Same as above, fromUtf8. Also would it maek sense to have an QString errorMessage(GError); internal method? or even let Response take a GError.

aleixpol (commented):
Would it be possible to include properties and invokables to make sure we can use this from QML eventually as well?

libportal-qt/filechooser.h
grulja: Options are content of comboboxes:
```
        <varlistentry>
          <term>choices a(ssa(ss)s)</term>
          <listitem>
            <para>
              List of serialized combo boxes to add to the file chooser.
            </para>
            <para>
              For each element, the first string is an ID that will be returned
              with the response, the second string is a user-visible label. The
              a(ss) is the list of choices, each being an ID and a
              user-visible label. The final string is the initial selection,
              or "", to let the portal decide which choice will be initially selected.
              None of the strings, except for the initial selection, should be empty.
            </para>
            <para>
              As a special case, passing an empty array for the list of choices
              indicates a boolean choice that is typically displayed as a check
              button, using "true" and "false" as the choices.
            </para>
            <para>
              Example: [('encoding', 'Encoding', [('utf8', 'Unicode (UTF-8)'), ('latin15', 'Western')], 'latin15'), ('reencode', 'Reencode', [], 'false')]
            </para>
          </listitem>
        </varlistentry>
      </variablelist>
```

grulja:
> Would it be possible to include properties and invokables to make sure we can use this from QML eventually as well?

I added Q_INVOKABLE to all portal calls, as well as Q_PROPERTY into the Response class, but I'm not sure about adding Q_PROPERTY into the rest of classes, like Notification or FileChooser[Filter,Choice] as I cannot imagine using those from QML. Instead, it might make sense to add some direct methods instead taking simple filters as strings and similar.

libportal-qt/filechooser.h
aleixpol: ok :+1:

aleixpol:
LGTM :+1:

grulja:
> > May I ask you to start with review of the current portal implementations while I keep finishing the remaining ones?
> 
> It might be a good idea to get someone familiar with Qt/KDE to review this, it's mostly client code.

There is also quite a lot of GVariant parsing code for which I would appreciate review from someone familiar with GLib.

hadess:
> There is also quite a lot of GVariant parsing code for which I would appreciate review from someone familiar with GLib.

I think your best bet here is to write tests. I don't know anyone who gets that stuff right first or third time.

You can use `g_variant_print()` to get a string you can reuse to create your test data.

grulja:
> > There is also quite a lot of GVariant parsing code for which I would appreciate review from someone familiar with GLib.
> 
> I think your best bet here is to write tests. I don't know anyone who gets that stuff right first or third time.
> 
> You can use `g_variant_print()` to get a string you can reuse to create your test data.

I verified everything myself using the Qt demo, but I will try to write tests for that as well to be sure.

grulja:
> I'm unable to make CI to build and run the tests, while running same commands locally I can run them just fine. Does anyone know what might be wrong? Maybe the older version of Meson is the reason?

Fixed!!

@hadess may I ask you for a review?

autotests/meson.build
```diff
@@ -0,0 +1,4 @@
+auto_tests = get_option('autotests')
```
GeorgesStavracas, line 1: Please rename this folder and meson property to `tests`

meson_options.txt
```diff
@@ -1,4 +1,11 @@
 option('portal-tests', type: 'array', choices: ['gtk3', 'gtk4', 'qt'], value: [],
   description : 'A list of portal tests to build')
+
+option('autotests', type: 'array', choices: ['qt'], value: [],
+description : 'A list of portal unit tests to build')
+
+option('libportal-qt', type: 'boolean', value: true,
```
GeorgesStavracas, line 7: Naming this option as `qt` or `qt5` reads more meson-y

meson_options.txt
```diff
@@ -1,4 +1,11 @@
 option('portal-tests', type: 'array', choices: ['gtk3', 'gtk4', 'qt'], value: [],
   description : 'A list of portal tests to build')
+
+option('autotests', type: 'array', choices: ['qt'], value: [],
+description : 'A list of portal unit tests to build')
```
GeorgesStavracas, line 5: I'd rather have this option as a boolean, and inside `tests/meson.build` you`d add something like:

```
if get_option('qt5')
  subdir('qt5')
endif
```

autotests/qt/meson.build
```diff
@@ -0,0 +1,21 @@
+
+add_languages('cpp', required : true)
+
+qt_dep = dependency('qt5', modules: ['Core', 'Test'])
```
GeorgesStavracas, line 4: What do you think of installing a separate pkg-config file for libportal-qt5?

GeorgesStavracas (changes requested):
I wonder why this PR does not include screencast-related portals... does Qt support exporting window handlers on Wayland?

autotests/qt/meson.build
grulja: I do install pkg-config file. See libportal-qt/meson.build and in the middle there is 

```
pkgconfig.generate(libportalqt,
  description: 'Qt portal API wrappers',
  name: 'libportal-qt',
  requires: [ qt5_dep ],
)
```

grulja:
> I wonder why this PR does not include screencast-related portals... does Qt support exporting window handlers on Wayland?

Qt doesn't support export window handlers on Wayland yet. 

The reason why this PR doesn't include other portals is that it took me quite a lot of time to implement the most common ones (lack of time) and I wanted to ship at least reasonable subset so I can start finally using it.

autotests/qt/meson.build
grulja: I now see what you meant with your question. We now have two pkg-config files since you install one and I install another one for the wrappet. It is slightly confusing and I think I should probably pick a different name for the Qt wrapper. I also think I should rename it to include Qt5 in the name, because one day there might be a Qt6 version.

grulja:
What do you think of this idea?

Drop most of the code I have here, since it duplicates the original code, and extend the Qt code we have with those additional classes from this PR (like Notification or FileChooserFilter) so we can just have easier way to generate GVariants we need to pass in some of the functions. I would also add CMake support to the Qt backend.

This would satisfy my needs and make it easier to use for C++/Qt devs, while avoiding to duplicate the code by providing just slightly different API.

GeorgesStavracas:
> extend the Qt code we have 

I assume you're talking about the new `libportal-qt5` library that is generated now? If so, I agree, moving this code there will be better :slightly_smiling_face: 

> I would also add CMake support to the Qt backend.

How is CMake related to this library? Is it about installing some sort of `FindLibportal.cmake` file? I don't think adding another build system here would be a good idea, but if it's about installing an extra file to help CMake, that's probably fine.

grulja:
> > extend the Qt code we have
> 
> I assume you're talking about the new `libportal-qt5` library that is generated now? If so, I agree, moving this code there will be better 
 
Yes, that's exactly what I mean. I will try to add some reasonable amount of code from this PR just to make it more convenient to use.

> > I would also add CMake support to the Qt backend.
> 
> How is CMake related to this library? Is it about installing some sort of `FindLibportal.cmake` file? I don't think adding another build system here would be a good idea, but if it's about installing an extra file to help CMake, that's probably fine.

Yes, it's about making the library to be found and usable from CMake without using pkg-config.

GeorgesStavracas:
Sounds like a plan then! Thanks for working on this, and apologies for stepping on your toes with libportal-qt5

hadess:
I had a bit of time so tried to review this, but I was completely overwhelmed by the size of the single commit. Can you please at least split adding the tests from the qt5 support? That would be really helpful.

grulja:
Closing in favor of https://github.com/flatpak/libportal/pull/69.

PR #52, opened 2021-11-07
Some build system cleanups

GeorgesStavracas:
See commits. No functional changes here (but we **are** compiling less targets now!).


GeorgesStavracas:
Thank you!

Issue #12, opened 2019-12-19
Make bindable -gtk3 and -gtk4 libraries


matthiasclasen:
While the inline header trick is cute, we should probably just install separate .so s for the benefit of bindings

PR #53, opened 2021-11-07
Add separate libraries for GTK3 / GTK4 / Qt 5

GeorgesStavracas:
Separate each XdpParent implementation into its own library, and install each lib's header under a different prefix (`libportal-{gtk3|gtk4|qt5}`)

I'm not entirely satisfied with the naming picked here - `backend` - but that's the best I could come up with. Suggestions appreciated.

Fixes https://github.com/flatpak/libportal/issues/12


TingPing (approved):
This all looks very good to me.

GeorgesStavracas:
Thanks again for the timely review

PR #55, opened 2021-11-08
Flatpak & CI improvements

GeorgesStavracas:
Generate Flatpak bundles for all manifests we have in-tree.

Cleanup and increment the builds to include:

 * Ubuntu 20.04
 * Ubuntu 21.10
 * Fedora 35

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,63 @@
+---
+
+name: PortalTest Flatpak
+
+on:
+  push:
+    paths-ignore: ['**.md']
+    branches: [master]
+  pull_request:
+    paths-ignore: ['**.md']
+    branches: [master]
+
+jobs:
+  build-flatpak-gtk3:
+    name: GTK3
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:gnome-41
+      options: --privileged
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build GTK3 portal test
+        uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v3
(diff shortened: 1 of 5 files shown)


GeorgesStavracas:
I'm not ecstatic about fighting Ubuntu 18.04 to check for the different backends, but at least all backends are compiled by the Flatpak jobs.

TingPing:
> I'm not ecstatic about fighting Ubuntu 18.04 to check for the different backends

I'm totally ok with adding newer Ubuntu, on top of that one, there.

```yaml
runs-on: ubuntu-latest
container: ubuntu:21.04
```

For example. (Note the apt commands need to be tweaked a bit, see https://github.com/flatpak/xdg-desktop-portal/blob/master/.github/workflows/check.yml)

.github/workflows/flatpak.yml
```diff
@@ -0,0 +1,63 @@
+---
+
+name: Flatpak
```
TingPing, line 3: Name this "Portal Test Flatpak", more clear what it is.

GeorgesStavracas: Sure

TingPing:
Note that 21.04 and 21.10 have GTK4 while 20.04 doesn't.

GeorgesStavracas:
This is roughly the limit I'm comfortable growing the `build.yml` file. If we want more distros, it might be better to switch to an OS matrix or any other fancy technique.

TingPing (approved):
Looks good.

TingPing:
Whats the plan for building the qt backend?

GeorgesStavracas:
> Whats the plan for building the qt backend?

I'm not touching anything Qt-related anymore until @grulja lands the improved `libportal-qt5` with more API coverage, etc.

PR #56, opened 2021-11-11
portal-tests/gtk4: Carefully parse streams

GeorgesStavracas:
Streams may not have (x, y), and we need to account for that.
This fixes stream sizes not showing up when selecting windows
to screencast on GNOME.


TingPing (approved):
Looks reasonable.

GeorgesStavracas:
Thanks

PR #57, opened 2021-11-15
build-aux/build.sh: Build all manual tests as Flatpak apps

smcv:
Previously this convenience script only built the Gtk3 version.


GeorgesStavracas (approved):
Looks good

PR #58, opened 2021-11-17
remote: Add support for VIRTUAL sources

GeorgesStavracas:
This is supported by the ScreenCast portal now, so reflect this in the API.


GeorgesStavracas:
CC @jadahl

GeorgesStavracas:
Thanks!

PR #59, opened 2021-11-18
Screencast session restore

GeorgesStavracas:
Add the XdpPersistMode enum, and modify the signature of
xdp_portal_create_screencast_session() to account for the
persist mode, and the restore token.

Add getters of the restore token and the persist mode to
XdpSession, and document that they are only available after
starting the session.

Adjust the corresponding code of portal tests. Make the GTK4
portal test exercise the transient persist mode.

### DIFF
--- a/libportal/portal-private.h
+++ b/libportal/portal-private.h
@@ -48,6 +48,9 @@ struct _XdpPortal {
 
   /* notification */
   guint action_invoked_signal;
+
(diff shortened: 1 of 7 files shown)


TingPing (approved):
LGTM

GeorgesStavracas:
Thank you!

PR #63, opened 2021-12-03
Add API for checking for flatpak/snap sandbox

mwleeds:
Applications need to know if they are running under a sandbox to know if
they should use portal APIs rather than direct access, in the cases that
the portal isn't built in to the toolkit and the portal isn't used by
both sandboxed and unsandboxed apps. Applications may also have other
reasons for wanting to know their sandbox status, such as Epiphany which
currently disables non-working web app functionality in that case.

Currently it's a wild west where each app has to have its own helper
function to check its sandbox status. For Flatpak one checks for the
existence of /.flatpak-info and for Snap one checks the process's
cgroups membership. The code for this (copied from xdg-desktop-portal)
is not completely trivial so it's good to have it shared between apps.
Possibly x-d-p itself should also change its libportal dependency to a
hard requirement and use the code here, but for that we would need to
allow checking by pid rather than only using the current process PID.

One might wonder if we should have these functions use a gboolean out
parameter and reserve the return value only for the success/failure, but
that would create a lot of tedium for users of the API that call it
often or as part of complex conditional statements. Also, these
functions really shouldn't fail under any normal circumstance, but if we
add API for checking these properties of another pid other than
getpid(), then errors would be more likely (e.g. permissions).

### DIFF
--- a/libportal/portal-helpers.h
+++ b/libportal/portal-helpers.h
@@ -33,6 +33,15 @@ XDP_PUBLIC
 G_DECLARE_FINAL_TYPE (XdpPortal, xdp_portal, XDP, PORTAL, GObject)
 
 XDP_PUBLIC
-XdpPortal *xdp_portal_new                    (void);
+XdpPortal *xdp_portal_new                   (void);
+
+XDP_PUBLIC
+gboolean   xdp_portal_running_under_flatpak (void);
+
+XDP_PUBLIC
+gboolean   xdp_portal_running_under_snap    (GError **error);
+
+XDP_PUBLIC
+gboolean   xdp_portal_running_under_sandbox (void);
 
 G_END_DECLS
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -23,6 +23,13 @@
 #include "portal-private.h"
 #include "portal-enums.h"
 
+#include <unistd.h>
+#include <string.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <sys/vfs.h>
+#include <stdio.h>
+
 /**
  * XdpPortal
  *
@@ -272,4 +279,159 @@ XdpPortal *
 xdp_portal_new (void)
 {
   return g_object_new (XDP_TYPE_PORTAL, NULL);
+}
+
+/* This function is copied from xdg-desktop-portal */
+int
+_xdp_parse_cgroup_file (FILE *f, gboolean *is_snap)
+{
+  ssize_t n;
+  g_autofree char *id = NULL;
+  g_autofree char *controller = NULL;
+  g_autofree char *cgroup = NULL;
+  size_t id_len = 0, controller_len = 0, cgroup_len = 0;
+
+  g_return_val_if_fail (f != NULL, -1);
+  g_return_val_if_fail (is_snap != NULL, -1);
+
+  *is_snap = FALSE;
+  do
+    {
+      n = getdelim (&id, &id_len, ':', f);
+      if (n == -1) break;
+      n = getdelim (&controller, &controller_len, ':', f);
+      if (n == -1) break;
+      n = getdelim (&cgroup, &cgroup_len, '\n', f);
+      if (n == -1) break;
+
+      /* Only consider the freezer, systemd group or unified cgroup
+       * hierarchies */
+      if ((!strcmp (controller, "freezer:") != 0 ||
+           !strcmp (controller, "name=systemd:") != 0 ||
+           !strcmp (controller, ":") != 0) &&
+          strstr (cgroup, "/snap.") != NULL)
+        {
+          *is_snap = TRUE;
+          break;
+        }
+    }
+  while (n >= 0);
+
+  if (n < 0 && !feof(f)) return -1;
+
+  return 0;
+}
+
+/* This function is copied from xdg-desktop-portal pid_is_snap() */
+static gpointer
+get_under_snap (gpointer user_data)
+{
+  g_autofree char *cgroup_path = NULL;;
+  int fd;
+  FILE *f = NULL;
+  gboolean is_snap = FALSE;
+  int err = 0;
+  pid_t pid = getpid ();
+  GError **error = user_data;
+
+  cgroup_path = g_strdup_printf ("/proc/%u/cgroup", (guint) pid);
+  fd = open (cgroup_path, O_RDONLY | O_CLOEXEC | O_NOCTTY);
+  if (fd == -1)
+    {
+      err = errno;
+      goto end;
+    }
+
+  f = fdopen (fd, "r");
+  if (f == NULL)
+    {
+      err = errno;
+      goto end;
+    }
+
+  fd = -1; /* fd is now owned by f */
+
+  if (_xdp_parse_cgroup_file (f, &is_snap) == -1)
+    err = errno;
+
+  fclose (f);
+
+end:
+  /* Silence ENOENT, treating it as "not a snap" */
+  if (err != 0 && err != ENOENT)
+    {
+      g_set_error (error, G_IO_ERROR, g_io_error_from_errno (err),
+                   "Could not parse cgroup info for pid %u: %s", (guint) pid,
+                   g_strerror (err));
+      return GINT_TO_POINTER (FALSE);
+    }
+  return GINT_TO_POINTER (is_snap);
+}
+
+/**
+ * xdp_portal_running_under_flatpak:
+ *
+ * Detects if running inside of a Flatpak or WebKit sandbox.
+ *
+ * See also: xdp_portal_running_under_sandbox()
+ * Returns: %TRUE if the current process is running under a Flatpak sandbox
+ */
+gboolean
+xdp_portal_running_under_flatpak (void)
+{
+  static gsize under_flatpak;
+
+  if (g_once_init_enter (&under_flatpak))
+    {
+      gboolean flatpak_info_exists = g_file_test ("/.flatpak-info", G_FILE_TEST_EXISTS);
+      g_once_init_leave (&under_flatpak, flatpak_info_exists);
+    }
+
+  return under_flatpak;
+}
+
+/**
+ * xdp_portal_running_under_snap:
+ * @error: return location for a #GError pointer
+ *
+ * Detects if you are running inside of a Snap sandbox.
+ *
+ * See also: xdp_portal_running_under_sandbox()
+ * Returns: %TRUE if the current process is running under a Snap sandbox, or
+ *   %FALSE if either unsandboxed or an error was encountered in which case
+ *   @error will be set
+ */
+gboolean
+xdp_portal_running_under_snap (GError **error)
+{
+  static GOnce under_snap_once = G_ONCE_INIT;
+  static GError *cached_error = NULL;
+  gboolean under_snap;
+
+  under_snap = GPOINTER_TO_INT (g_once (&under_snap_once, get_under_snap, &cached_error));
+
+  if (error != NULL && cached_error != NULL)
+    g_propagate_error (error, g_error_copy (cached_error));
+
+  return under_snap;
+}
+
+/**
+ * xdp_portal_running_under_sandbox:
+ *
+ * This function tries to determine if the current process is running under a
+ * sandbox that requires the use of portals.
+ *
+ * If you need to check error conditions see xdp_portal_running_under_snap().
+ *
+ * Note that these functions are all cached and will always return the same result.
+ *
+ * Returns: %TRUE if the current process should use portals to access resources
+ *   on the host system, or %FALSE if either an error was encountered or the
+ *   process is running unsandboxed
+ */
+gboolean
+xdp_portal_running_under_sandbox (void)
+{
+  return xdp_portal_running_under_flatpak () || xdp_portal_running_under_snap (NULL);
 }


libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+
+/**
+ * xdp_portal_running_under_flatpak:
+ * @error: return location for a #GError pointer
+ *
+ * If you are trying to determine whether to use portals, consider using
+ * xdp_portal_running_under_sandbox() instead.
+ *
+ * Returns: %TRUE if the current process is running under a Flatpak sandbox, or
+ *   %FALSE if either an error was encountered or the process is running
```
TingPing, line 292: All of these only return an error one time, following calls will be `FALSE` without error which is unclear.

libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+  static _Thread_local gboolean under_flatpak = FALSE;
+  g_autofree char *root_path = NULL;
+  int root_fd = -1;
+  int info_fd = -1;
+
+  if (decided)
+    return under_flatpak;
+
+  /* Note: the rest of this function is copied from xdg-desktop-portal */
+  root_path = g_strdup_printf ("/proc/%u/root", getpid ());
```
TingPing, line 308: This complication seems to exist entirely for discovering if *other* processes are running under flatpak. Surely the task is quite a bit easier when checking in your own root and much of this can be simplified?

mwleeds: Well, it doesn't seem impossible that the current process is running on a fuse root filesystem, which is the main source of the complexity. So I guess it depends whether we value avoiding complexity with a simple `g_file_test ("/.flatpak-info", G_FILE_TEST_EXISTS);` more than not setting the `GError` in that case.

libportal/portal.c
mwleeds: hmm, should we just print warnings on the error code paths? That would still only be once per process but that behavior seems okay.

libportal/portal.c
mwleeds: However experimentally `/proc/getpid()/root` is readable from a process running under toolbox, so maybe you're right and we can get rid of a lot of this.

libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+
+      /* Otherwise, we should be able to open the root dir. Probably the app died and
+         we're failing due to /proc/$pid not existing. In that case fail instead
+         of treating this as privileged. */
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Unable to open %s", root_path);
+      goto end;
+    }
+
+  info_fd = openat (root_fd, ".flatpak-info", O_RDONLY | O_CLOEXEC | O_NOCTTY);
```
sonnyp, line 334: > As a side note, the /.flatpak-info mechanism for identifying a process as untrusted to a portal is perhaps not ideal. [...] Perhaps an area for improvement if anyone has ideas?

Since Flatpak 1.1.2 you can use `$FLATPAK_ID`.

https://github.com/flatpak/flatpak/releases/tag/1.1.2

libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+gboolean
+xdp_portal_running_under_snap (GError **error)
+{
+  static _Thread_local gboolean decided = FALSE;
+  static _Thread_local gboolean under_snap = FALSE;
+
+  if (decided)
+    return under_snap;
+
+  under_snap = pid_is_snap (getpid (), error);
```
sonnyp, line 463: Why not using an environment var here as well?

Allows for testing code paths easily and seems much simpler than this.

https://snapcraft.io/docs/environment-variables

mwleeds: It would definitely be simpler but strictly speaking less reliable, because a thread could clear its environment variables right?

libportal/portal.c
mwleeds: Good to know but I don't think we can assume Flatpak is at least version 1.1.2, and that is only a possibility when checking your own Flatpak'd status, not that of another process, so it's not a replacement for the `/.flatpak-info` mechanism.

mcatanzaro: > Since Flatpak 1.1.2 you can use $FLATPAK_ID.

This won't work for WebKit's flatpak-like sandbox, which only creates .flatpak-info and needs to be supported as well.

libportal/portal.c
mwleeds: I pushed a fixup that should address this.

libportal/portal.c
mwleeds: I changed this to have separate API depending on whether its your own pid or another pid.

libportal/portal-helpers.h
```diff
@@ -35,6 +35,27 @@ XDP_PUBLIC
 G_DECLARE_FINAL_TYPE (XdpPortal, xdp_portal, XDP, PORTAL, GObject)
 
 XDP_PUBLIC
-XdpPortal *xdp_portal_new                    (void);
+XdpPortal *xdp_portal_new                       (void);
+
+XDP_PUBLIC
+gboolean   xdp_portal_pid_running_under_flatpak (pid_t    pid,
```
TingPing, line 41: Who is the consumer of these pid APIs?

I'm just not sure any project, other than the portals themselves, care about this in practice? And I just think this library is for client applications rather than portal services.

mwleeds: I was thinking xdg-desktop-portal but it looks like there is at least one other potential user because Tracker has also copied [this code](https://github.com/GNOME/tracker/blob/5e612b2e55d569b8e5b228a124cf59a36d200e3d/src/portal/tracker-portal-utils.c#L156) from x-d-p. But tracker reads the key file so we'd need to add that here if we want to support that use case.

Maybe there are other system components that need to know whether to treat a process as trusted. You think it's better to get rid of the pid APIs here?

TingPing: Yes, I think we should only add public APIs that have clear users. Since it also can greatly simplify the code removing it, I think its best to remove it.

libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+ *
+ * Returns: %TRUE if the current process is running under a Flatpak sandbox, or
+ *   %FALSE if either an error was encountered or the process is running
+ *   unsandboxed
+ */
+gboolean
+xdp_portal_running_under_flatpak (GError **error)
+{
+  static _Thread_local gboolean decided = FALSE;
+  static _Thread_local gboolean under_flatpak = FALSE;
```
GeorgesStavracas: You could use `GPrivate` for thread-specific data.

libportal/portal.c
```diff
@@ -273,3 +280,200 @@ xdp_portal_new (void)
...
+ *
+ * Returns: %TRUE if the current process is running under a Snap sandbox, or
+ *   %FALSE if either an error was encountered or the process is running
+ *   unsandboxed
+ */
+gboolean
+xdp_portal_running_under_snap (GError **error)
+{
+  static _Thread_local gboolean decided = FALSE;
+  static _Thread_local gboolean under_snap = FALSE;
```
GeorgesStavracas, line 523: Ditto about `GPrivate`

libportal/portal.c
GeorgesStavracas: I think avoiding environment variables is the correct decision to make here.

libportal/portal-helpers.h
GeorgesStavracas: The original goal for libportal is to be used by applications, not other services. This PID-based API did seem a bit out of place to me too.

GeorgesStavracas (changes requested):
Some comments and suggestions

libportal/portal.c
mwleeds: Now changed it to use the simpler `g_file_test` check

libportal/portal.c
mwleeds: Is there an advantage to doing so? I copied `_Thread_local` from epiphany's codebase, and the `GPrivate` documentation says it is "a very limited resource", so I don't know what the reason to switch would be.

libportal/portal-helpers.h
mwleeds: Alright, I removed the pid APIs.

libportal/portal.c
mcatanzaro: I think `GPrivate` predates `_Thread_local`. They're surely the same thing.

libportal/portal.c
```diff
@@ -273,3 +280,159 @@ xdp_portal_new (void)
...
+    return under_flatpak;
+
+  decided = TRUE;
+  under_flatpak = g_file_test ("/.flatpak-info", G_FILE_TEST_EXISTS);
+  return under_flatpak;
+}
+
+/**
+ * xdp_portal_running_under_snap:
+ * @error: return location for a #GError pointer
```
TingPing, line 394: Still the same comment about the error here. Maybe just keep a copy of it and set error every call if it failed.

libportal/portal.c
```diff
@@ -273,3 +280,159 @@ xdp_portal_new (void)
...
+ *
+ * If you are trying to determine whether to use portals, consider using
+ * xdp_portal_running_under_sandbox() instead.
+ *
+ * Returns: %TRUE if the current process is running under a Flatpak sandbox
+ */
+gboolean
+xdp_portal_running_under_flatpak (void)
+{
+  static _Thread_local gboolean decided = FALSE;
```
TingPing, line 381: You can include `<threads.h` and have a nicer looking `thread_local`.

Also you should add checks to the build system for this to have a nice error in case somebody uses a non-c11 compiler.

libportal/portal.c
TingPing: You can include `<threads.h>` and have a nicer looking `thread_local`.

Also you should add checks to the build system for this to have a nice error in case somebody uses a non-c11 compiler.

TingPing: (I kinda think depending on C11 for this is a bit silly, but its also not new...)

TingPing: I'm actually not sure this should be thread-local also. Like these do blocking IO in the first place so its not like waiting on a mutex is a problem, and we want every thread to have the exact same result every time.

mcatanzaro: The only point of thread local is to avoid mutex locking. So if you're doing I/O anyway, there's indeed no point: just use a GOnce.

libportal/portal.c
mwleeds: Don't think that's necessary. It already only sets `decided = TRUE` if there was no error.

libportal/portal.c
mwleeds: I don't understand what makes `GOnce` better than `_Thread_local` in this case?

mcatanzaro: `_Thread_local` is a precious resource that has to be used sparingly.

You originally copied this from Epiphany, where it doesn't matter so much, but now this is a library that will be used in loads of applications.

GeorgesStavracas:
@mcatanzaro regarding https://github.com/flatpak/libportal/pull/63#discussion_r762447787, could you please enlighten this ignorant mind as for why WebKit tries to pretend it is a Flatpak sandbox? What's the rationaly behind that? Do WebKit subprocesses use portals?

TingPing:
> @mcatanzaro regarding #63 (comment), could you please enlighten this ignorant mind as for why WebKit tries to pretend it is a Flatpak sandbox? What's the rationaly behind that? Do WebKit subprocesses use portals?

Yes it uses portals, it has its own sandbox. `xdg-desktop-portal` had one way of checking if a process is sandbox and that was by looking for `/.flatpak-info`. So WebKit simply creates that file. WebKit *could* have made a `/.webkit-info` and then updated `xdg-desktop-portal` but this would have no practical benefits, it would have the same contents, and all tooling already checks for the other file.

libportal/portal.c
mwleeds: And `GOnce` will let the check run again if an unsandboxed process forks off a process that is sandboxed? Maybe that's a dumb question because I don't have a good grasp on how bubblewrapped processes are created for rendering thumbnails or whatever else they're used for.

TingPing: GOnce is simply a mutex protected global variable. It is only set once in a process. Subprocesses won't be affected.

mwleeds: Ok, I updated it to use GOnce. It became a bit more complex but maybe it's fine.

mwleeds:
In theory this is ready to go except I haven't tested it so we should do that before merging.

libportal/portal.c
TingPing: As I mentioned earlier I think `g_once_init_enter()` is the cleaner method.

Sorry for the back and forth.

mwleeds: I don't think that would work for `xdp_portal_running_under_snap` since we call the `get_under_snap` function in the `SNAP_STATE_ERROR` case even if it's been called before?

mwleeds:
> In theory this is ready to go except I haven't tested it so we should do that before merging.

Finally managed to test this under both snap and flatpak and it works as expected.

libportal/portal.c
```diff
@@ -273,3 +280,171 @@ xdp_portal_new (void)
...
+    {
+      case SNAP_STATE_UNSANDBOXED:
+        return FALSE;
+      case SNAP_STATE_SANDBOXED:
+        return TRUE;
+      case SNAP_STATE_ERROR:
+        if (error == NULL)
+          return FALSE;
+        else
+          return GPOINTER_TO_INT (get_under_snap (error)) == SNAP_STATE_SANDBOXED;
```
TingPing, line 428: This is no longer thread safe.

mwleeds: I don't see how anything bad would happen if two threads are executing `get_under_snap` at the same time?

TingPing: Ok maybe the wrong term but the current function does weird things:


1. Only run get_under_snap() once in one thread
2. But actually if it errored run it more times in many threads and don't cache the result

TingPing: I've pushed two new commits. Let me know if they look OK to you.

libportal/portal.c
```diff
@@ -273,3 +280,158 @@ xdp_portal_new (void)
...
+ * See also: xdp_portal_running_under_sandbox()
+ * Returns: %TRUE if the current process is running under a Snap sandbox, or
+ *   %FALSE if either unsandboxed or an error was encountered in which case
+ *   @error will be set
+ */
+gboolean
+xdp_portal_running_under_snap (GError **error)
+{
+  static GOnce under_snap_once = G_ONCE_INIT;
+  static GError *cached_error = NULL;
```
mwleeds, line 408: We're leaking this memory when this pointer is set right?

TingPing: Yes. We could use `g_atexit()` but I've heard that is problematic for libs.

libportal/portal.c
mwleeds: Squashed them in. If we're ok with the `GError` memory leak I think this can be merged.

libportal/portal.c
mcatanzaro: I think it's OK because you have it assigned to a static variable. Memory leak checkers should not complain about a one-time allocation here.

TingPing:
Merged with minor tweaks as f95a1e1a275fed22985f130cbb6a302c94cb4799

PR #65, opened 2021-12-10
WIP: Generate vapi

A6GibKm:


A6GibKm:
The gtk4 vapi file is empty.

libportal/meson.build
```diff
@@ -138,7 +148,19 @@ if 'gtk3' in backends
       link_with: [libportal_gtk3],
       includes: [libportal_gir[0], 'Gtk-3.0'],
       install: true,
+      extra_args: ['--c-include=libportal-gtk3/portal-gtk3.h'],
+      export_packages: ['libportal-gtk3'],
      )
+
+    if vapi
+      libportal_gtk3_vapi = gnome.generate_vapi('libportal-gtk3',
```
GeorgesStavracas, line 156: This variable is not used anywhere, I don't think you need it. (Same for the GTK4 variable)

libportal/meson.build
```diff
@@ -138,7 +148,19 @@ if 'gtk3' in backends
...
+        sources: libportal_gtk3_gir[0],
+        packages: ['gio-2.0', 'gtk+-3.0', 'libportal'],
+        gir_dirs: [meson.current_build_dir()],
+        vapi_dirs: [meson.current_build_dir()],
+        install: true
```
GeorgesStavracas, line 161: Nit: comma in this line too (same for the GTK4 part)

GeorgesStavracas (changes requested):

libportal/meson.build
nahuelwexd: If any Vala project wants to include libportal as a subproject, it will need this variable, otherwise valac will not be able to find the corresponding vapi

A6GibKm: I added `libportal_vapi` just in case.

PR #70, opened 2022-01-07
camera: Use correct variant for is_camera_present()

nielsdg:
The `org.freedesktop.DBus.Properties.Get()` method always returns a
variant of type `(v)`, no matter what the actual type of the property
is.

Before this commit, `xdp_portal_is_camera_present()` was broken, as it
outputted CRITICALs on each call, since the variant type was wrong.

### DIFF
--- a/libportal/camera.c
+++ b/libportal/camera.c
@@ -37,7 +37,7 @@ xdp_portal_is_camera_present (XdpPortal *portal)
 {
   g_autoptr(GError) error = NULL;
   g_autoptr(GVariant) ret = NULL;
-  gboolean result;
+  g_autoptr(GVariant) prop = NULL;
 
   g_return_val_if_fail (XDP_IS_PORTAL (portal), FALSE);
 
@@ -47,7 +47,7 @@ xdp_portal_is_camera_present (XdpPortal *portal)
                                      "org.freedesktop.DBus.Properties",
                                      "Get",
                                      g_variant_new ("(ss)", "org.freedesktop.portal.Camera", "IsCameraPresent"),
-                                     G_VARIANT_TYPE_BOOLEAN,
+                                     G_VARIANT_TYPE ("(v)"),
                                      G_DBUS_CALL_FLAGS_NONE,
                                      -1,
                                      NULL,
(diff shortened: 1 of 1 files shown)


libportal/camera.c
```diff
@@ -47,7 +48,7 @@ xdp_portal_is_camera_present (XdpPortal *portal)
                                      "org.freedesktop.DBus.Properties",
                                      "Get",
                                      g_variant_new ("(ss)", "org.freedesktop.portal.Camera", "IsCameraPresent"),
-                                     G_VARIANT_TYPE_BOOLEAN,
+                                     G_VARIANT_TYPE ("(v)"),
```
GeorgesStavracas, line 51:
```suggestion
                                     G_VARIANT_TYPE_VARIANT,
```

libportal/camera.c
```diff
@@ -58,7 +59,8 @@ xdp_portal_is_camera_present (XdpPortal *portal)
       return FALSE;
     }
 
-  g_variant_get (ret, "(b)", &result);
+  g_variant_get (ret, "(v)", &prop);
+  result = g_variant_get_boolean (prop);
```
GeorgesStavracas, line 63: You can remove the `result` variable now, and return `g_variant_get_boolean (prop)` directly

GeorgesStavracas (changes requested):
Looks fine. I still root for this portal to be retired in favor of the Devices portal, but it is what it is.

libportal/camera.c
nielsdg: Isn't that different though `"(v)"` vs `"v"`?

libportal/camera.c
nielsdg: done

libportal/camera.c
smcv: @nielsdg is correct. The `reply_type` for a method call in GDBus is always a tuple with zero or more types in parentheses, never just a plain type (because replies can "return" more than one "out" parameter).

In this case we're representing

```
o.fd.DBus.Properties.Get(s: interface, s: property) -> v: value
```

as

```
in parameters: (ss)
out parameters: (v)
```

smcv (approved):

libportal/camera.c
GeorgesStavracas: I see, thanks for clarifying

GeorgesStavracas (approved):
LGTM

PR #72, opened 2022-01-12
Fixes location and updates signal emissions

A6GibKm:


portal-test/gtk4/window.ui
```diff
@@ -564,6 +564,50 @@
...
+          <object class="GtkLabel">
+            <property name="label">Latitude:</property>
+            <layout>
+              <property name="column">2</property>
+              <property name="row">20</property>
+            </layout>
+          </object>
+        </child>
+        <child>
+          <object class="GtkLabel" id="lat">
```
GeorgesStavracas, line 587: `latitudeLabel`

libportal/location.c
```diff
@@ -185,6 +185,8 @@ call_returned (GObject *object,
       g_task_return_error (call->task, error);
       create_call_free (call);
     }
+  else
+    ensure_location_updated_connected (call->portal);
```
GeorgesStavracas, line 189:
```suggestion
    {
      ensure_location_updated_connected (call->portal);
    }
```

libportal/updates.c
```diff
@@ -343,7 +342,10 @@ update_started (GObject *object,
   if (error)
     g_task_return_error (call->task, error);
   else
-    g_task_return_boolean (call->task, TRUE);
+   {
+     ensure_update_monitor_connection (call->portal);
+     g_task_return_boolean (call->task, TRUE);
+   }
```
GeorgesStavracas, line 348:
```c

  if (error)
    {
      g_task_return_error (call->task, error);
    }
  else
   {
     ensure_update_monitor_connection (call->portal);
     g_task_return_boolean (call->task, TRUE);
   }
```

GeorgesStavracas (changes requested):

smcv:
> If done too soon it won't work

In what way does it not work?

I'm cautious about this because delaying a D-Bus signal connection until after a successful method call often leads to this anti-pattern:

* client: call GetThing()
* server: reply GetThing() -> 23
* server: emit ThingChanged(to: 42)
* client: receive reply GetThing() -> 23
* client: subscribe to ThingChanged
* too late! client will never find out that Thing changed to 42

A6GibKm:
> > If done too soon it won't work
> 
> In what way does it not work?
> 
> I'm cautious about this because delaying a D-Bus signal connection until after a successful method call often leads to this anti-pattern:
> 
>     * client: call GetThing()
> 
>     * server: reply GetThing() -> 23
> 
>     * server: emit ThingChanged(to: 42)
> 
>     * client: receive reply GetThing() -> 23
> 
>     * client: subscribe to ThingChanged
> 
>     * too late! client will never find out that Thing changed to 42

It does not work in the sense that LocationUpdated will never be emitted on the other side of the DBus.

A6GibKm:
I am not convinced this is the solution, but before this changes the location portal would fail about 19/20 times, now it only fails about 1/8 of times.

A6GibKm:
I think ill close it, it does not work on my other machine. This issue is the weirdest condition race I have ever seen. 

All I know this issue is in libportal side since, ASHPD works in a reliable way.way.

A6GibKm:
I updated the MR so just to include the location portal test to the demo.

libportal/updates.c
GeorgesStavracas: Neither this one

PR #75, opened 2022-01-20
Avoid a critical in the sandbox check code

mwleeds:
Passing FALSE (zero) to g_once_init_leave() isn't allowed and leads to
GLib-CRITICAL **: 22:02:50.836: g_once_init_leave: assertion 'result !=0' failed

So rework xdp_portal_running_under_flatpak() a bit.

### DIFF
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -379,7 +379,7 @@ end:
 gboolean
 xdp_portal_running_under_flatpak (void)
 {
-  static gsize under_flatpak;
+  static gsize under_flatpak = 0;
   enum {
     NOT_FLATPAK = 1,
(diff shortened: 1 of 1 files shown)


mwleeds:
(I did test this even though it's pretty trivial)

libportal/portal.c
```diff
@@ -380,14 +380,21 @@ gboolean
 xdp_portal_running_under_flatpak (void)
 {
   static gsize under_flatpak;
```
smcv, line 382:
```suggestion
  static gsize under_flatpak = 0;
```

because explicit is better than implicit?

Other than that, LGTM

PR #78, opened 2022-01-28
Add dynamic launcher portal

mwleeds:
The corresponding x-d-p PR is here:
https://github.com/flatpak/xdg-desktop-portal/pull/696


libportal/dynamic-launcher.c
```diff
@@ -0,0 +1,609 @@
...
+ * This function gets the icon associated with a .desktop file that was previously
+ * installed by the dynamic launcher portal.
+ *
+ * The @desktop_file_id must be prefixed with the caller's app ID followed by a
+ * "." and suffixed with ".desktop".
+ *
+ * Returns: (transfer full): the icon in a format recognized by g_icon_deserialize(),
+ *   or %NULL with @error set
+ */
+GVariant *
```
GeorgesStavracas, line 526: Shouldn't it return a `GIcon*` instead?

libportal/dynamic-launcher.c
```diff
@@ -0,0 +1,609 @@
...
+ * launcher to their desktop.
+ *
+ * When the request is done, @callback will be called. You can then
+ * call [method@Portal.dynamic_launcher_prepare_install_finish] to get the results.
+ */
+void
+xdp_portal_dynamic_launcher_prepare_install (XdpPortal           *portal,
+                                             XdpParent           *parent,
+                                             const char          *name,
+                                             GVariant            *icon_v,
```
GeorgesStavracas, line 251: Ditto

GeorgesStavracas:
I don't have much else to say, this LGTM once we figure out the `GIcon*` situation

libportal/dynamic-launcher.c
mwleeds: I guess it depends how closely we want the libportal API to match the underlying portal API, but I don't see much benefit. Using `GIcon` here and in the `xdp_portal_dynamic_launcher_prepare_install` signature means the caller doesn't have to do `g_icon_serialize`/`g_icon_deserialize`, but it also means libportal has to handle erroring out if the `GIcon` is not a `GBytesIcon` (or let the portal error out instead). Doesn't really seem like enough of a benefit to justify reworking this and the calling code?

mwleeds: >it also means libportal has to handle erroring out if the GIcon is not a GBytesIcon (or let the portal error out instead)

I guess we can just replace `g_return_val_if_fail (g_variant_is_of_type (icon_v, G_VARIANT_TYPE ("(sv)")), NULL);` with `g_return_val_if_fail (G_IS_BYTES_ICON (icon), NULL);` when the `GIcon` is being passed in. But then we have to handle setting the error pointer if serialization fails (GLib doesn't set it for us). And similarly we would need to handle setting the error pointer if deserialization fails in `xdp_portal_dynamic_launcher_get_icon()`.

I don't see any use of `g_set_error()` in libportal code aside from in `get_under_snap()`, so I don't know if we want to start using it and make an error domain for libportal (or use something generic like `G_IO_ERROR_FAILED` but that's an anti-pattern). Seems like that would make it a bit harder to reason about when client code is checking error conditions, since the status quo is that errors are only set by portal implementations or at the layer of gdbus, I think.

Another reason it might not make sense to use `GIcon` is in case some clients would want the icon in serialized form when retrieving it with `xdp_portal_dynamic_launcher_get_icon()` or if some clients would already have it serialized and have to deserialize it to pass it as a `GIcon`.

Just some thoughts, not a strong opinion.

GeorgesStavracas: If this code requires a specific implementation of GIcon, then perhaps it should use that as the accepted type. E.g. `GBytesIcon * xdp_portal_dynamic_launcher_... ()`. Though, on a second thought, using the implementation type directly would limit any future API changes.

mwleeds: Perhaps we should stick with `GVariant` then?

GeorgesStavracas: Let's go with GVariant for now

GeorgesStavracas (approved):
Please rebase, squash, and land it

PR #85, opened 2022-03-22
ci: Add ABI check job

hadess:


hadess:
Are the jobs in the `build.yml` file not run? This is my first time modifying a github workflows definition...

hadess:
> Are the jobs in the `build.yml` file not run? This is my first time modifying a github workflows definition...

All sorted (the jobs don't appear straight away in the list of jobs to check, so I ended up with 3/3 from the Flatpak job showing at first)

.github/workflows/build.yml
```diff
@@ -88,3 +88,23 @@ jobs:
           publish_dir: ./_build/doc/libportal-1/
           cname: libportal.org
           destination_dir: ./
+
+  fedora-35-abi:
+    name: Fedora 35 ABI check
```
GeorgesStavracas, line 93: Will the ABI checker run differently depending on the distro version? If not, I'd recommend calling it `ABI check` only

GeorgesStavracas (commented):
One minor comment, looks good otherwise

GeorgesStavracas (approved):

PR #88, opened 2022-05-05
libportal: Add activation-token to notification-action-invoked

3v1n0:
Since some time FDO Notifications introduced the ActivationToken signal
to ensure that clients will receive an X11-startup notification ID or a
wayland xdg-activation token in order to support stealing-focus
prevention protocols.

However this isn't supported by xdg-desktop-portal's.
So fill the gap by adding a further activation token.

While it could be added by as an optional parameter, it's just better to
make it explicit.

Required by https://github.com/flatpak/xdg-desktop-portal/pull/791


3v1n0:
@GeorgesStavracas have some time for this (and maybe the related portal one)? :smile:

jsparber:
The user can obtain an activation token since notification v2 is now merged. https://github.com/flatpak/xdg-desktop-portal/pull/1298 I think this can be closed.

PR #96, opened 2022-06-02
Add support for the InputCapture portal

whot:
For the corresponding Portal PR, please see https://github.com/flatpak/xdg-desktop-portal/pull/714

This is a draft PR to add support for the input capture portal. Most of it is straightforward portal implementation but this portal does have a larger surface and more state than most, which raises a few questions:

- [ ] ~~`XdpSession` is re-used for input capture portals now too. This makes namespacing (or the lack thereof) slightly awkward, e.g. the `activated` signal on the session is instead `inputcapture-activated` to avoid future conflicts, which then requires the other signals to be namespaced too.~~ 
  - ~~ideally we'd have a `XdpInputCaptureSession` as subclass of `XdpSession` but the latter is declared final right now and I'm not 100% if that's an ABI break then.~~ edit: this was nacked in #102 
- [ ] `XdpInputCaptureSession` now *contains* an `XdpSession` and the API is namespaced properly.
- [x] The `enable`/`disable`/`release` methods ~~are paired with their respective `_finish`, but tbh this is of questionable value. Even if the portal keeps these as requests, I'd argue hiding this away here would make for better API - we can use the session signals to notify the caller out-of-band.~~ Update June 06: these are method calls now
- [ ] the pointer barriers and capture zones have their own GObjects with properties and little other functionality. This makes for slightly more sensible API but it also conflicts a bit with e.g. `XdpSession` which doesn't have properties but setters/getters only.
- [x] the `Disabled` signal is still missing here I just noticed
- [x] there's a dbusmock/pytest based test suite included. ~~that won't run without patched dbus-python and dbusmock~~ but it makes testing *so* much easier. specifically, it allows to test libportal without an existing implementation of the actual portal. Merged to main in #99, this is just the inputcapture tests now

The above are the larger topics, I'd appreciate a high-level review of the API and suggestions for where to go with this.

cc @jadahl


whot:
if anyone is following along: the most recent changes from today change the implementation so that we have an API that creates and takes `XdpInputCaptureSession` objects. Inside that object is an `XdpSession` object that represents the session, it's exposed to the user on request via `xdp_input_capture_session_get_session()` so the user can e.g. subscribe to `closed`.

This is basically a working implementation of #112 (while not touching the RD/SC API) and should eventually allow us to create an InputCapture session from an existing XdpSession object, see https://github.com/flatpak/xdg-desktop-portal/pull/864.

libportal/inputcapture.h
```diff
@@ -0,0 +1,101 @@
...
+                                                     XdpParent            *parent,
+                                                     XdpCapability         capabilities,
+                                                     GCancellable         *cancellable,
+                                                     GAsyncReadyCallback   callback,
+                                                     gpointer              data);
+
+XDP_PUBLIC
+XdpInputCaptureSession * xdp_portal_create_input_capture_session_finish (XdpPortal    *portal,
+                                                                         GAsyncResult *result,
+                                                                         GError       **error);
```
jadahl, line 65:
```suggestion:-2
XdpInputCaptureSession * xdp_portal_create_input_capture_session_finish (XdpPortal     *portal,
                                                                         GAsyncResult  *result,
                                                                         GError       **error);
```

libportal/inputcapture-pointerbarrier.c
```diff
@@ -0,0 +1,275 @@
...
+  PROP_Y2,
+  PROP_ID,
+  PROP_ACTIVE,
+
+  N_PROPERTIES
+};
+
+enum
+{
+  SIGNAL_ACTIVE,
```
jadahl, line 55: This isn't needed since `PROP_ACTIVE` already provides this via `notify::active`.

libportal/inputcapture.c
```diff
@@ -0,0 +1,1170 @@
...
+  XdpInputCaptureSession *session = XDP_INPUT_CAPTURE_SESSION (data);
+  g_autoptr(GVariant) options = NULL;
+  const char *handle = NULL;
+
+  g_variant_get(parameters, "(o@a{sv})", &handle, &options);
+
+  if (!handle_matches_session (session, handle))
+    return;
+
+  g_signal_emit_by_name (session, "activated", options);
```
jadahl, line 413: Can use `g_signal_emit()` with the enum directly. Same elsewhere.

libportal/inputcapture-pointerbarrier.c
```diff
@@ -0,0 +1,275 @@
...
+                        0, UINT_MAX, 0,
+                        G_PARAM_CONSTRUCT_ONLY | G_PARAM_READWRITE);
+  /**
+   * XdpInputCapturePointerBarrier:is-active:
+   *
+   * A boolean indicating whether this barrier is active. A barrier cannot
+   * become active once it failed to apply, barriers that are not active can
+   * be thus cleaned up by the caller.
+   */
+  properties[PROP_ACTIVE] =
```
jadahl, line 243: `PROP_IS_ACTIVE` to match the name.

libportal/inputcapture-pointerbarrier.c
```diff
@@ -0,0 +1,275 @@
...
+   * XdpInputCapturePointerBarrier:id:
+   *
+   * The caller-assigned unique id of this barrier
+   */
+  properties[PROP_ID] =
+    g_param_spec_uint ("id",
+                        "Pointer barrier unique id",
+                        "The id assigned to this barrier by the caller",
+                        0, UINT_MAX, 0,
+                        G_PARAM_CONSTRUCT_ONLY | G_PARAM_READWRITE);
```
jadahl, line 235: Indentation is off (below too).

libportal/inputcapture-private.h
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: LGPL-3.0-only
+ */
+
+#include "inputcapture-pointerbarrier.h"
+
+int
+_xdp_input_capture_pointer_barrier_get_id (XdpInputCapturePointerBarrier *barrier);
+
+void
+_xdp_input_capture_pointer_barrier_set_active (XdpInputCapturePointerBarrier *barrier, gboolean active);
```
jadahl, line 26: set_is_active() to match the property name.

jadahl (commented):
Some light review so far only, looks good to me so far.

libportal/inputcapture-pointerbarrier.c
whot: indeed, thanks. leftover from a previous version and removed now. All other comments fixed as requested.

whot:
Just FTR, this now sits on top of #114 because I need both to work at the same time and juggling branches is annoying.

whot:
marking as ready, I stared at this for too long, I can no longer see any bugs...

Conan-Kudo:
What's keeping this from landing at this point? It seems to have been sitting in a ready state for a month now...

libportal/inputcapture.h
```diff
@@ -0,0 +1,103 @@
...
+
+G_BEGIN_DECLS
+
+#define XDP_TYPE_INPUT_CAPTURE_SESSION (xdp_input_capture_session_get_type ())
+
+XDP_PUBLIC
+G_DECLARE_FINAL_TYPE (XdpInputCaptureSession, xdp_input_capture_session, XDP, INPUT_CAPTURE_SESSION, GObject)
+
+/**
+ * XdpCapability:
```
GeorgesStavracas, line 36: How about XdpInputCapability? Or anything else to namespace this enum

GeorgesStavracas (approved):
One small comment. LGTM otherwise.

libportal/inputcapture-zone.c
```diff
@@ -0,0 +1,245 @@
...
+
+struct  _XdpInputCaptureZone {
+    GObject parent_instance;
+
+    guint width;
+    guint height;
+    gint x;
+    gint y;
+
+    guint zone_set;
```
ebassi, line 60:
```suggestion
    unsigned int width;
    unsigned int height;
    int x;
    int y;

    unsigned int zone_set;
```

Same as above.

libportal/inputcapture-pointerbarrier.c
```diff
@@ -0,0 +1,255 @@
...
+    enum barrier_state state;
+};
+
+G_DEFINE_TYPE (XdpInputCapturePointerBarrier, xdp_input_capture_pointer_barrier, G_TYPE_OBJECT)
+
+static void
+xdp_input_capture_pointer_barrier_finalize (GObject *object)
+{
+  G_OBJECT_CLASS (xdp_input_capture_pointer_barrier_parent_class)->finalize (object);
+}
```
ebassi, line 83: If the implementation is empty, there's no need to override.

libportal/inputcapture.c
```diff
@@ -0,0 +1,1207 @@
...
+  XdpSession *parent_session; /* strong ref */
+
+  GList *zones;
+
+  guint signal_ids[SIGNAL_LAST_SIGNAL];
+  guint zone_serial;
+  guint zone_set;
+};
+
+G_DEFINE_TYPE (XdpInputCaptureSession, xdp_input_capture_session, G_TYPE_OBJECT)
```
ebassi, line 73:
```suggestion
G_DEFINE_FINAL_TYPE (XdpInputCaptureSession, xdp_input_capture_session, G_TYPE_OBJECT)
```

This way, people will get a warning if they try subclassing the type by using the instance/class sizes known to the type system.

libportal/inputcapture-zone.c
```diff
@@ -0,0 +1,245 @@
...
+
+    gboolean is_valid;
+};
+
+G_DEFINE_TYPE (XdpInputCaptureZone, xdp_input_capture_zone, G_TYPE_OBJECT)
```
ebassi, line 65:
```suggestion
G_DEFINE_FINAL_TYPE (XdpInputCaptureZone, xdp_input_capture_zone, G_TYPE_OBJECT)
```

This way, people using the type system to query the instance and class sizes in order to subclass this type will hit an assertion failure.

libportal/inputcapture.c
```diff
@@ -0,0 +1,1207 @@
...
+    g_signal_handler_disconnect (g_task_get_cancellable (call->task), call->cancelled_id);
+
+  g_free (call->request_path);
+
+  if (call->portal)
+    g_object_unref (call->portal);
+  if (call->task)
+    g_object_unref (call->task);
+  if (call->session)
+    g_object_unref (call->session);
```
ebassi, line 266:
```suggestion
  g_clear_object (&call->portal);
  g_clear_object (&call->task);
  g_clear_object (&call->session);
```

Even if there's no re-entrancy, using `g_clear_object()` is more future-proof, and it adds no real cost.

libportal/inputcapture.c
```diff
@@ -0,0 +1,1207 @@
...
+  g_autoptr(GVariant) ret;
+
+  ret = g_dbus_connection_call_finish (G_DBUS_CONNECTION (object), result, &error);
+  if (error)
+    {
+      if (call->cancelled_id)
+        {
+          g_signal_handler_disconnect (g_task_get_cancellable (call->task), call->cancelled_id);
+          call->cancelled_id = 0;
+        }
```
ebassi, line 289:
```suggestion
      g_clear_signal_handler (&call->cancelled_id, g_task_get_cancellable (call->task));
```

libportal/inputcapture.c
```diff
@@ -0,0 +1,1207 @@
...
+          gpointer         data)
+{
+  XdpInputCaptureSession *session = XDP_INPUT_CAPTURE_SESSION (data);
+  g_autoptr(GVariant) options = NULL;
+  const char *handle = NULL;
+
+  g_variant_get(parameters, "(o@a{sv})", &handle, &options);
+
+  if (!handle_matches_session (session, handle))
+          return;
```
ebassi, line 469:
```suggestion
  if (!handle_matches_session (session, handle))
    return;
```

Coding style.

matthiasclasen:
I had this on my list to look at today. Thanks for beating me to it, @ebassi

libportal/inputcapture.h
ofourdan: Done.

libportal/inputcapture-pointerbarrier.c
```diff
@@ -0,0 +1,255 @@
...
+};
+
+static GParamSpec *properties[N_PROPERTIES] = { NULL, };
+
+struct  _XdpInputCapturePointerBarrier {
+    GObject parent_instance;
+
+    guint id;
+    gint x1, y1;
+    gint x2, y2;
```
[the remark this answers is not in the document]
ofourdan: Sure, done.

I also changed `guint` to `unsigned int` for the `property_id` in the `get_property()/set_property()` handlers for consistency.

libportal/inputcapture-zone.c
```diff
@@ -0,0 +1,245 @@
...
+
+static void
+xdp_input_capture_zone_finalize (GObject *object)
+{
+  G_OBJECT_CLASS (xdp_input_capture_zone_parent_class)->finalize (object);
+}
```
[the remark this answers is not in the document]
ofourdan: Removed!

libportal/session.c
```diff
@@ -50,6 +50,7 @@ static void
 xdp_session_finalize (GObject *object)
 {
   XdpSession *session = XDP_SESSION (object);
+  g_warning ("session finalize");
```
ofourdan, line 53: Spurious warning.

libportal/inputcapture-pointerbarrier.c
ofourdan: Removed as well.

libportal/inputcapture.c
ofourdan: Done.

libportal/inputcapture-zone.c
ofourdan: Done as well.

libportal/inputcapture.c
ofourdan: Done.

libportal/inputcapture.c
ofourdan: Indentation fixed.

libportal/session.c
ofourdan: Removed

libportal/inputcapture.c
ofourdan: Undone, sorry, that breaks the build (CI) with Ubuntu 20.04.

libportal/inputcapture-zone.c
ofourdan: Undone, sorry, that breaks the build (CI) with Ubuntu 20.04.

libportal/inputcapture.c
ofourdan: Undone, sorry, that breaks the build (CI) with Ubuntu 20.04.

ofourdan:
Many thanks for the review!

I have applied the changes and forced push to @whot's branch now.

Please also notethat I had to drop a few suggested chnages at it was braking the build on Ubuntu 20.04 for the CI (I suspect the version of Glib is older on that target).

whot:
ftr, I fixed a buggy test case, but this branch should still be ready to go.

whot:
gentle ping for merging :)

GeorgesStavracas (approved):

PR #99, opened 2022-06-17
test: add a pytest/dbusmock-based test suite

whot:
Using python and dbusmock makes it trivial to add a large number of
tests for libportal only, without requiring an actual portal
implementation for the Portal interface to be tested.

Included here is the wallpaper portal as an example, hooked into meson test.
A helper script is provided too for those lacking `meson devenv`,
```
  $ ./test/gir-testenv.sh
  $ cd test
  $ pytest --verbose --log-level=DEBUG [... other pytest arguments ...]
```
The test setup uses dbusmock interface templates (see
pyportaltest/templates) to handle the actual DBus calls.

Because DBus uses a singleton for the session bus, we need libportal to
specifically connect to the address given in the environment - otherwise
starting mock dbus services has no effect.

This test suite depends on [dbusmock commit 4a191d8ba293:
"mockobject: allow sending signals with extra details"](https://github.com/martinpitt/python-dbusmock/pull/129)

Without this, the `EmitSignalDetailed()` method does not exist/work, but
without this method we cannot receive signals.

### DIFF
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -254,12 +254,47 @@ xdp_portal_class_init (XdpPortalClass *klass)
                   G_TYPE_VARIANT);
 }
 
+static GDBusConnection *
+create_bus_from_address (const char *address,
+                         GError    **error)
+{
+  g_autoptr(GDBusConnection) bus = NULL;
+
+  if (!address)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Missing D-Bus session bus address");
+      return NULL;
+    }
+
+  bus = g_dbus_connection_new_for_address_sync (address,
+                                                G_DBUS_CONNECTION_FLAGS_AUTHENTICATION_CLIENT |
+                                                G_DBUS_CONNECTION_FLAGS_MESSAGE_BUS_CONNECTION,
+                                                NULL, NULL,
+                                                error);
+  return g_steal_pointer (&bus);
+}
+
 static void
 xdp_portal_init (XdpPortal *portal)
 {
+  g_autoptr(GError) error = NULL;
   int i;
 
-  portal->bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, NULL);
+  /* g_bus_get_sync() returns a singleton. In the test suite we may restart
+   * the session bus, so we have to manually connect to the new bus */
+  if (getenv ("LIBPORTAL_TEST_SUITE"))
+    portal->bus = create_bus_from_address (getenv ("DBUS_SESSION_BUS_ADDRESS"), &error);
+  else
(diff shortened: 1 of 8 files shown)


libportal/portal.c
```diff
@@ -259,7 +259,27 @@ xdp_portal_init (XdpPortal *portal)
...
+                                                            G_DBUS_CONNECTION_FLAGS_AUTHENTICATION_CLIENT|G_DBUS_CONNECTION_FLAGS_MESSAGE_BUS_CONNECTION,
+                                                            NULL, NULL, &error);
+      if (error)
+        {
+          g_error("Failed to connect to bus: %s\n", error->message);
+          abort();
+        }
+    }
+  else
+    portal->bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, NULL);
```
GeorgesStavracas, line 282: I think this would look a bit less convoluted if you split the first part into a separate function, something like this (notice the subtle code style suggestions too!):

```c
static GDBusConnection *
create_bus_from_address (const char  *address,
                         GError     **error)
{
  g_autoptr(GDBusConnection) bus = NULL;

  if (!address)
    {
      // Please think of a better message :)
      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Invalid D-Bus address");
      return NULL;
    }

  bus = g_dbus_connection_new_for_address_sync (address,
                                                G_DBUS_CONNECTION_FLAGS_AUTHENTICATION_CLIENT |
                                                G_DBUS_CONNECTION_FLAGS_MESSAGE_BUS_CONNECTION,
                                                NULL, NULL,
                                                &error);

  return g_steal_pointer (&bus);
}

...

  if (getenv ("LIBPORTAL_TEST_SUITE"))
    portal->bus = create_bus_from_address (getenv ("DBUS_SESSION_BUS_ADDRESS", &error);
  else
    portal->bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &error);

  if (error)
    {
      g_critical ("Could not create XdpPortal instance: %s", error->message);
      abort ();
    }

  g_assert (portal->bus != NULL);
```

GeorgesStavracas (changes requested):

libportal/portal.c
whot: Done, thanks. Though I admit my error message is not exactly the pinnacle of creativity either...

whot:
Just ftr, I squashed in the changes requested above together with a bit of cleanup in the tests to bring them slightly closer to how https://github.com/flatpak/xdg-desktop-portal/pull/764 looks atm. Frustratingly, all these test suites are very close but not the same  so they can't be fully re-used (the Xdp one provides the impl, this one here provides the portal.

GeorgesStavracas (approved):

PR #102, opened 2022-06-28
Subclass RemoteDesktop/ScreenCast sessions from XdpSession

whot:
`XdpSession` is currently a mixed struct, comprising of both the bits required for a Session in general and the bits required for a ScreenCast session and/or a RemoteDesktop Session (the latter is *usually* a ScreenCast session too but does not have to be).

This works though only because we effectively only have one real subtype. With InputCapture using sessions as well we're getting namespace crowding and collisions. Fix this by switching `XdpSession` to be a derivable object and then subclass `XdpScreenCastSession` from it and then `XdpRemoteDesktopSession` from that.

API-wise nothing changes, the returned objects are now of the right subclass but the API itself remains unchanged. Ideally we could deprecate the current ones and switch the others to "type-safe" APIs but that's mostly busywork and I can't come up with a non-clashing nice enough name for the `create_...` calls.

cc @jadahl

### DIFF
--- a/libportal/remote.c
+++ b/libportal/remote.c
@@ -41,6 +41,139 @@ typedef struct {
   guint cancelled_id;
 } CreateCall;
 
+struct _XdpScreenCastSessionClass
+{
+  XdpSessionClass parent_class;
+};
+
+typedef struct {
+  GVariant *streams;
+  XdpPersistMode persist_mode;
+  char *restore_token;
+} XdpScreenCastSessionPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (XdpScreenCastSession, xdp_screencast_session, XDP_TYPE_SESSION)
+
+struct _XdpRemoteDesktopSessionClass
+{
+  XdpScreenCastSessionClass parent_class;
+};
+
+typedef struct {
+  XdpDeviceType devices;
+  gboolean is_screencast;
+} XdpRemoteDesktopSessionPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (XdpRemoteDesktopSession, xdp_remotedesktop_session, XDP_TYPE_SCREENCAST_SESSION)
+
+
+static void
+xdp_screencast_session_finalize (GObject *object)
+{
+  XdpScreenCastSession *session = XDP_SCREENCAST_SESSION (object);
+  XdpScreenCastSessionPrivate *priv = xdp_screencast_session_get_instance_private (session);
+
+  g_clear_pointer (&priv->restore_token, g_free);
+  g_clear_pointer (&priv->streams, g_variant_unref);
+
+  G_OBJECT_CLASS (xdp_screencast_session_parent_class)->finalize (object);
+}
+
+static void
+xdp_screencast_session_class_init (XdpScreenCastSessionClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->finalize = xdp_screencast_session_finalize;
+}
+
+static void
+xdp_screencast_session_init (XdpScreenCastSession *session)
+{
+}
+
+static void
+xdp_remotedesktop_session_class_init (XdpRemoteDesktopSessionClass *klass)
+{
+}
+
+static void
+xdp_remotedesktop_session_init (XdpRemoteDesktopSession *session)
+{
+}
+
+/**
+ * xdp_session_get_streams:
+ * @session: a [class@Session]
+ *
+ * Obtains the streams that the user selected.
+ *
+ * The information in the returned [struct@GLib.Variant] has the format
+ * `a(ua{sv})`. Each item in the array is describing a stream. The first member
+ * is the pipewire node ID, the second is a dictionary of stream properties,
+ * including:
+ *
+ * - position, `(ii)`: a tuple consisting of the position `(x, y)` in the compositor
+ *     coordinate space. Note that the position may not be equivalent to a
+ *     position in a pixel coordinate space. Only available for monitor streams.
+ * - size, `(ii)`: a tuple consisting of (width, height). The size represents the size
+ *     of the stream as it is displayed in the compositor coordinate space.
+ *     Note that this size may not be equivalent to a size in a pixel coordinate
+ *     space. The size may differ from the size of the stream.
+ *
+ * Unless the session is active, this function returns `NULL`.
+ *
+ * Returns: the selected streams
+ */
+GVariant *
+xdp_session_get_streams (XdpSession *session)
+{
+  XdpScreenCastSession *sc; 
+  XdpScreenCastSessionPrivate *priv; 
+
+  g_return_val_if_fail (XDP_IS_SCREENCAST_SESSION (session), NULL);
+
+  if (xdp_session_get_session_state(session) != XDP_SESSION_ACTIVE)
+    return NULL;
+
+  if (XDP_IS_REMOTEDESKTOP_SESSION (session))
+    {
+      XdpRemoteDesktopSession *rd = XDP_REMOTEDESKTOP_SESSION (session); 
+      XdpRemoteDesktopSessionPrivate *rd_priv = xdp_remotedesktop_session_get_instance_private (rd);
+
+      g_return_val_if_fail (rd_priv->is_screencast, NULL);
+    }
+
+  sc = XDP_SCREENCAST_SESSION (session);
+  priv = xdp_screencast_session_get_instance_private (sc);
+
+  return priv->streams;
+}
+
+/**
+ * xdp_session_get_devices:
+ * @session: a [class@Session]
+ *
+ * Obtains the devices that the user selected.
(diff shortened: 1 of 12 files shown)


libportal/session.c
```diff
@@ -44,20 +44,40 @@ enum {
...
+    g_dbus_connection_signal_unsubscribe (priv->portal->bus, priv->signal_id);
 
-  g_clear_object (&session->portal);
-  g_free (session->restore_token);
-  g_free (session->id);
-  g_clear_pointer (&session->streams, g_variant_unref);
+  g_clear_object (&priv->portal);
+  g_free (priv->restore_token);
+  g_free (priv->id);
+  g_clear_pointer (&priv->streams, g_variant_unref);
```
GeorgesStavracas: Could you add an extra commit to this branch changing all these cleanup free()s and whatnot to `g_clear_*()`?

libportal/remote.c
```diff
@@ -746,7 +887,7 @@ xdp_session_start (XdpSession *session,
 
   call = g_new0 (StartCall, 1);
   call->portal = g_object_ref (_xdp_session_get_portal (session));
-  call->session = g_object_ref (session);
+  call->session = g_object_ref (XDP_SCREENCAST_SESSION (session));
```
GeorgesStavracas, line 890: Do you really need to cast here? `g_object_ref()` returns `gpointer` which shouldn't trigger any warnings

libportal/remote.c
```diff
@@ -195,13 +332,17 @@ devices_selected (GDBusConnection *bus,
         select_sources (call);
       else
         {
+          XdpRemoteDesktopSession *session = g_object_new (XDP_TYPE_REMOTEDESKTOP_SESSION, NULL);
+
```
GeorgesStavracas, line 336: One pattern I've been trying to adhere is to always declare GObject variables as `g_autoptr()`, and if necessary return / pass them using `g_steal_pointer()`.

libportal/remote.c
```diff
@@ -41,6 +41,132 @@ typedef struct {
...
+static void
+xdp_screencast_session_init (XdpScreenCastSession *session)
+{
+}
+
+static void
+xdp_remotedesktop_session_finalize (GObject *object)
+{
+  G_OBJECT_CLASS (xdp_remotedesktop_session_parent_class)->finalize (object);
+}
```
GeorgesStavracas, line 93: Seems like you don't need to override `finalize()` here?

libportal/remote.c
```diff
@@ -41,6 +41,132 @@ typedef struct {
...
+G_DEFINE_TYPE (XdpRemoteDesktopSession, xdp_remotedesktop_session, XDP_TYPE_SCREENCAST_SESSION)
+
+
+static void
+xdp_screencast_session_finalize (GObject *object)
+{
+  XdpScreenCastSession *session = XDP_SCREENCAST_SESSION (object);
+  XdpScreenCastSessionPrivate *priv = xdp_screencast_session_get_instance_private (session);
+
+  g_free (priv->restore_token);
```
GeorgesStavracas, line 70: Since this is "new" code, please switch to `g_clear_pointer()`

GeorgesStavracas (changes requested):
Standing ovation for the test included in this pull request

whot (commented):
All comments addressed as requested, thanks for the review!

libportal/remote.h
```diff
@@ -24,6 +24,25 @@
...
+G_DECLARE_DERIVABLE_TYPE (XdpScreenCastSession, xdp_screencast_session, XDP, SCREENCAST_SESSION, XdpSession)
+
+struct _XdpScreenCastSessionClass
+{
+  GObjectClass parent_class;
+
+  void (*close) (XdpSession *session);
+
+  gpointer padding[24];
+};
```
GeorgesStavracas, line 39: Seems like this needs to go to a `remote-private.h` header too? We don't want client application code deriving from it

GeorgesStavracas (changes requested):
A couple more comments

libportal/remote.h
whot: just ftr, this is resolved, it just required the `XdpRemoteDesktopSession` to be a private implementation too.

whot:
Any chance we get can get this merged? I have other branches depending on this (inputcapture in particular)

libportal/remote.c
```diff
@@ -846,6 +846,16 @@ xdp_session_open_pipewire_remote (XdpSession *session)
   return g_unix_fd_list_get (fd_list, fd_out, NULL);
 }
 
+static gboolean
+is_active_remotedesktop_session (XdpSession    *session,
```
jadahl, line 983: nit: should be `is_active_remote_desktop_session()` to be consistent with other API.

libportal/session-private.h
```diff
@@ -21,20 +21,13 @@
...
-  XdpSessionState state;
-  XdpDeviceType devices;
-  GVariant *streams;
+  void (*close) (XdpSession *session);
 
-  XdpPersistMode persist_mode;
-  char *restore_token;
-
-  guint signal_id;
+  gpointer padding[24];
```
jadahl, line 30: Is this padding really needed with this sitting in a private header?

libportal/remote.c
```diff
@@ -41,6 +41,139 @@ typedef struct {
...
+  XdpSessionClass parent_class;
+};
+
+typedef struct {
+  GVariant *streams;
+  XdpPersistMode persist_mode;
+  char *restore_token;
+} XdpScreenCastSessionPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (XdpScreenCastSession, xdp_screencast_session, XDP_TYPE_SESSION)
```
jadahl, line 55: nit: it's `screen_cast` and `remote_desktop` elsewhere, not `screencast` and `remotedesktop`. Might cause confusion if we mix and match.

jadahl (commented):
For the screencast/remote desktop case, this seems doable, but what do we do about other "ad hoc" capabilities, would such be added?

Example I have is https://github.com/flatpak/xdg-desktop-portal/pull/852 which adds `org.freedesktop.portal.Clipboard` as another API that is used on the same session as the remote desktop session. The intention is to potentially allow reusing that API in other portals, to avoid duplication. With that in mind, it's not possible to fit `Clipboard` into some kind of subclass anymore.

Any thoughts about that? Or maybe the approach of https://github.com/flatpak/xdg-desktop-portal/pull/852 is wrong?

whot:
> For the screencast/remote desktop case, this seems doable, but what do we do about other "ad hoc" capabilities, would such be added?

I would argue that a more future-proof approach would be session bundling rather than just extending existing session implementations. SC/RD can be grandfathered in because they're too intertwined.

I'd argue for that primarily because it'll become impossible to figure out the right combinations of interactions - iirc there's one call in screencast/remotedesktop that can switch the session from one to the other, but that call needs to happen at the right session state. In the Clipboard case here, does `Start` affect what you can do? If not, then they're independent, if so then, whoah, we're in for some spec writing pain if we start adding InputCapture too :)

Maybe we should just have a generic `org.freedesktop.portal.Session.Create()` instead of per-portal ones and then use the session as-is? AFAICT nothing happens on session creation anyway, all the interaction is later when we have portal-specific calls.
And with "bundling" I mean - some way of combining several independent sessions into one logical entity that can be Closed as one?

jadahl:
> I would argue that a more future-proof approach would be session bundling rather than just extending existing session implementations.

Mh, not so sure about that, the point of sharing the session is that it's a single "handle" for an activity that was granted permission for; by making them two, it allows for them to work independently (except if we say they can't, but well, why have them separate if they can't be used separately?). For example, in the RD/SC, you set up your expectations for a session, then "start" it, and in response to the start call, a dialog handling all the things is presented. How do you deal with that if you have to start 2 sessions for the sake of what should really be one? "Bundling" them sounds like it'd still be a single session, containing two sessions, so in SC/RD, we went from 1 to 2, or 3 if the "bundled" session is a "parent" session, and adding clipboard, we end up with 3 (or 4).

Maybe it's just wrong to think of sessions as being tied to an "object' ala object oriented programming, and more about some "thing" that the user asked for special permissions for, that was granted.

> Maybe we should just have a generic org.freedesktop.portal.Session.Create() instead of per-portal ones and then use the session as-is?

That'd work too, IMO, the SC/RD `CreateSession` doesn't really do anything anyway. It'd kind of go in another direction than this PR though.

whot:
> Maybe it's just wrong to think of sessions as being tied to an "object' ala object oriented programming, and more about some "thing" that the user asked for special permissions for, that was granted.

There's a Draft API proposal in #112 now on how this could look like, together with https://github.com/flatpak/xdg-desktop-portal/pull/864

whot:
Closing this one, this is clearly a dead end.

#112 has the initial draft of how detangling SC/RD from the session namespace looks like and #96 has the implementation for `InputCapture`, it's already using the new XdpSession-inside-other-session approach.

PR #103, opened 2022-06-30
Up the warning level and fix the resulting warnings

whot:
cc @mwleeds for the cgroup hierarchies, `!strcmp() != 0` was sufficiently confusing that I'm going to punt this to you :smile: but based on the comment above, `g_str_equal()` should be the correct replacement.

### DIFF
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -339,9 +339,10 @@ _xdp_parse_cgroup_file (FILE *f, gboolean *is_snap)
 
       /* Only consider the freezer, systemd group or unified cgroup
        * hierarchies */
-      if ((!strcmp (controller, "freezer:") != 0 ||
-           !strcmp (controller, "name=systemd:") != 0 ||
-           !strcmp (controller, ":") != 0) &&
+      if (controller != NULL &&
+          (g_str_equal (controller, "freezer:") ||
+           g_str_equal (controller, "name=systemd:") ||
+           g_str_equal (controller, ":")) &&
           strstr (cgroup, "/snap.") != NULL)
         {
           *is_snap = TRUE;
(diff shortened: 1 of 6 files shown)


GeorgesStavracas (dismissed):

libportal/portal.c
```diff
@@ -339,9 +339,9 @@ _xdp_parse_cgroup_file (FILE *f, gboolean *is_snap)
 
       /* Only consider the freezer, systemd group or unified cgroup
        * hierarchies */
-      if ((!strcmp (controller, "freezer:") != 0 ||
-           !strcmp (controller, "name=systemd:") != 0 ||
-           !strcmp (controller, ":") != 0) &&
+      if ((g_str_equal (controller, "freezer:") ||
+           g_str_equal (controller, "name=systemd:") ||
+           g_str_equal (controller, ":")) &&
```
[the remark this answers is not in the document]
whot: the way the code flows, I think so. where a controller is not set, `getdelim()` would return -1 and we'd never get here. But it's simple enough to add a NULL-check so I've squashed that in now. Thanks for the review!

libportal/spawn.c
```diff
@@ -131,6 +131,8 @@ do_spawn (SpawnCall *call)
 
   ensure_spawn_exited_connection (call->portal);
 
+  g_variant_builder_init (&opt_builder, G_VARIANT_TYPE ("a{sv}"));
```
mwleeds, line 134:
```suggestion
  g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
```
for consistency with the rest of the codebase

mwleeds:
LGTM aside from the one thing I pointed out

libportal/spawn.c
whot: fixed

Issue #94, opened 2022-05-30
Omit qt / qt5 while compiling?


rubyFeedback:
Hey guys,

I am trying to compile libportal right now.

It tells me this though:

"qmake found: NO need ['>= 5', '< 6']"
"Run-time dependency qt5 (modules: Core, Gui, Widgets, X11Extras) found: NO (tried pkgconfig and config-tool)"

I tried to compile qt5 with the kde dev patchsets, but had a problem with wayland so I gave up for now.

I need libportal for nautilus and evince eventually. Is there a flag to skip building qt? If there is, would
it be possible to mention it in the main README briefly? Perhaps a new entry right in the existing
"Passing options" would suffice. I am not that familiar with meson and I am not sure how the equivalent
to "./configure --help" goes with meson. Thanks for reading either way.

PS: Indepedent of this, perhaps qt5 should not be a fail-error but just a warning. I believe libportal
can work fine without qt but for some reason it aborts right now since I have no qt installed. I think
the better default would be to continue building, UNLESS qt is mandatory for libportal.

rubyFeedback:
I just compiled libportal-0.4 just fine, so it seems as if some option changed in this regard between 0.4 and 0.5 (and/or perhaps with meson). libportal-0.4. compiled fine here.

eli-schwartz:
> I am not that familiar with meson and I am not sure how the equivalent
> to "./configure --help" goes with meson.

I can confirm the equivalent. The command you are looking for is:

```
meson configure
```

eli-schwartz:
Looks like you have the following build option: `-Dbackends=gtk3,gtk4,qt5`

Whichever ones are added to that list are a fail-error.

Whichever ones are not in that option list, are not a warning -- they are simply not checked for and not built.

eli-schwartz:
From the README:

> After first running meson _build, you can view the available options with:
> ```
> meson configure _build
> ```

This is incorrect, the README should be fixed. You do NOT need to run `meson _build` before viewing the available options. Those instructions are for viewing *configured* options, not *available* ones.

PR #104, opened 2022-07-01
meson: default the backends list to 'auto'

whot:
This changes the 'backends' option to the special string 'auto' - all
backends become optional by default and are built if the dependencies
are available.

If need be this can be combined with explicitly listing a backend, so
backends=auto,gtk3 will require the GTK3 backend and optionally build
the Qt5 and GTK4 backends. A summary is printed to list the actually
enabled backends and a warning where no backends are enabled at all.

This makes life easier for new users since the majority of them will not
have GTK3, GTK4 and Qt5 headers installed by default and thus are almost
guaranteed to run into a meson error on first configure.

Fixes https://github.com/flatpak/libportal/issues/94

### DIFF
--- a/libportal/meson.build
+++ b/libportal/meson.build
@@ -109,12 +109,14 @@ if introspection
   endif
 endif
 
+enabled_backends = []
+
 ########
 # GTK3 #
 ########
 
-if 'gtk3' in backends
-  gtk3_dep = dependency('gtk+-3.0')
+gtk3_dep = dependency('gtk+-3.0', required: get_option('backend-gtk3'))
+if gtk3_dep.found()
   gtk3_headers = ['portal-gtk3.h']
   gtk3_sources = ['portal-gtk3.c']
 
@@ -164,14 +166,15 @@ if 'gtk3' in backends
       )
     endif
   endif
+  enabled_backends += ['gtk3']
 endif
 
 ########
 # GTK4 #
 ########
 
-if 'gtk4' in backends
-  gtk4_dep = dependency('gtk4')
+gtk4_dep = dependency('gtk4', required: get_option('backend-gtk4'))
+if gtk4_dep.found()
   gtk4_headers = ['portal-gtk4.h']
   gtk4_sources = ['portal-gtk4.c']
 
@@ -222,16 +225,17 @@ if 'gtk4' in backends
       )
     endif
   endif
+  enabled_backends += ['gtk4']
 endif
 
 ########
 # Qt 5 #
 ########
 
-if 'qt5' in backends
-  add_languages('cpp', required : true)
+have_cpp = add_languages('cpp', required: get_option('backend-qt5'))
+qt5_dep = dependency('qt5', modules: ['Core', 'Gui', 'X11Extras', 'Widgets'], required: get_option('backend-qt5'))
 
-  qt5_dep = dependency('qt5', modules: ['Core', 'Gui', 'X11Extras', 'Widgets'])
+if have_cpp and qt5_dep.found()
   qt5_headers = ['portal-qt5.h']
   qt5_sources = ['portal-qt5.cpp']
 
@@ -257,4 +261,9 @@ if 'qt5' in backends
     dependencies: [libportal_dep, qt5_dep],
     link_with: libportal_qt5,
   )
+  enabled_backends += ['qt5']
+endif
+
+if meson.version().version_compare('>= 0.54.0')
+  summary({'enabled backends': enabled_backends}, section: 'Backends', list_sep: ',')
 endif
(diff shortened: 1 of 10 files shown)


whot:
cc @eli-schwartz for the inevitable meson magic that already exists for this and that I'm not aware of ;)

libportal/meson.build
```diff
@@ -109,12 +109,17 @@ if introspection
...
 # GTK3 #
 ########
 
-if 'gtk3' in backends
-  gtk3_dep = dependency('gtk+-3.0')
+require_gtk3 = 'gtk3' in backends
+want_gtk3 = 'auto' in backends or require_gtk3
+
+gtk3_dep = dependency('gtk+-3.0', required: require_gtk3)
+if want_gtk3 and gtk3_dep.found()
```
eli-schwartz, line 122: This will do a lookup for gtk3 even if the value is `qt5` and not auto. So, perhaps:

```meson
gtk3_dep = dependency('', required: false)

if want_gtk3
    gtk3_dep = dependency('gtk+-3.0', required: require_gtk3)
endif
if gtk3_dep.found()
    ...
```

libportal/meson.build
```diff
@@ -222,16 +232,25 @@ if 'gtk4' in backends
...
+want_qt5 = 'auto' in backends or require_qt5
+have_cpp = false
+
+if want_qt5
+  have_cpp = add_languages('cpp', required: require_qt5)
+endif
+
+qt5_dep = dependency('qt5', modules: ['Core', 'Gui', 'X11Extras', 'Widgets'], required: require_qt5)
+if want_qt5 and have_cpp and qt5_dep.found()
   add_languages('cpp', required : true)
```
eli-schwartz, line 252: You now have add_languages being done twice, but it's redundant here.

eli-schwartz (commented):
I dunno about "magic", but...

libportal/meson.build
whot: does the lookup matter much? Personally I'm in the camp of simpler code over lookups but if it comes at a cost other than the print line and a tiny amount of time, I can change this.

libportal/meson.build
whot: fixed, thanks!

libportal/meson.build
eli-schwartz: Well, even the time can be non-negligible considering that if gtk3 is not installed, meson will look up cmake, then run a cmake build in order to check if cmake can find it using a `find_package()` that will also eventually fail.

This can be confusing to users, apparently, due to mentioning cmake. ;) Why buy the trouble for a situation you know you will ignore it?

On the other hand you can fix that with a different code complexity. Specify the `method: ` kwarg.

...

Really, the general solution is feature options. Changing the array option into three different feature options may not be 100% desirable though... there have been discussions about instantiating custom feature objects, something like `make_feature_option('gtk3', required: require_gtk3, auto: want_gtk3)` except with a much better name and API obviously.

That's the magic I wish I could suggest.

whot: Changed as suggested!

whot:
gentle ping

whot:
gentle ping again

whot:
monthly ping again

whot:
@GeorgesStavracas - any chance we can merge this? Not having to manually supply the backends to build libportal from git is a rather nice feature.

whot:
ping again?

GeorgesStavracas:
@eli-schwartz I'd actually like to hear your thoughts on the "auto" magic that's happening here. I'm not convinced this is a great idea. Or, at least, I'd love to hear what a Mesonic way to implement this would look like.

GeorgesStavracas:
@whot sorry, completely dropped the ball here

whot:
My guess would be that the mesonic way would be to have each backend as separate feature option (instead of a "backends" array with a special string). That'd be easy enough to introduce but changes the build invocations.

eli-schwartz:
Yeah, adding a separate option for each one would be the most obvious way to handle it.

If you want to keep the existing option, I guess you could make each option default to auto, and then use something like:

```meson
backends = get_option('backends')
x_opt = 'x' in backends ? true : get_option('x')
y_opt = 'y' in backends ? true : get_option('y')
```

The option will then usually be a feature option, but will be a boolean true (equal to enabled when used inside `required:` kwargs) if the existing option specifies to build that backend. Check for the dependencies using `required: x_opt`, and they will use:
- auto /check by default
- the feature value specifically passed for that option
- true / enabled if listed in backends

Whichever one matches last.

whot:
Changed this PR to use `-Dbackend-gtk3=enabled` and so on. I tried keeping `backends` as list for backwards compatibility but tbh the conditions got confusing enough (in particular: `-Dbackends=gtk3` must also set the others to disabled) that I didn't really see the value in keeping it, so it's removed now.

GeorgesStavracas (approved):

PR #107, opened 2022-08-10
Fix build on FreeBSD

arrowd:


smcv (changes requested):
If libportal is useful on non-Linux (is it?), it would seem better to do this as a generic header check:

```
# meson.build
check_headers = ['sys/vfs.h']

foreach header : check_headers
    macro = 'HAVE_' + header.underscorify().to_upper()
    conf.set(macro, cc.has_header(header, args: compile_args_c) ? 1 : false)
endforeach
```                                                                       

```
/* portal.c */
#ifdef HAVE_SYS_VFS_H
#include <sys/vfs.h>
#endif
```

so that we'll include the header if and only if the platform has it. That would make sure the right thing happens on non-Linux, non-FreeBSD platforms, like the other BSDs.

smcv:
In general, "if the platform is not x, then do y" is a code smell, unless *x* is a platform that is known to be uniquely strange or deficient (usually Windows).

Where possible, it's usually better to have "if feature y exists, then use it".

Or, if it's necessary to do platform-based feature gates because only certain implementations are known to work, then it's usually better the other way round: "if the platform is x, then do y". For example, only using a Linux-specific thing like abstract sockets, `SO_PEERCRED`, systemd or user namespaces if the platform is Linux, so that for all other platforms (FreeBSD, NetBSD, Solaris, etc.) it won't be used, without needing a special case per non-Linux platform.

arrowd:
@smcv Applied your suggestion, thanks for the sample code!

smcv (approved):

PR #111, opened 2022-08-18
A bunch of cleanups in the session/testing code

whot:
Some of these commits come from #102

PR #112, opened 2022-08-19
Draft: Detangle Session creation from the RemoteDesktop/ScreenCast portals

whot:
Motivated by the discussion in  #102. This is on top of #111 with some commits from #102, ignore all but the top commit 

"Detangle Session creation from the RemoteDesktop/ScreenCast portals".

This is merely the propsed API, and only for RD/SC, Location is missing but would look similar.

This requires xdg-desktop-portal
- `portal.SessionFactory`, for `SessionFactory.CreateSession()`
- version 2 of `portal.RemoteDesktop`, for session_handle in `RemoteDesktop.CreateSession`
- version 5 of `portal.ScreenCast`, for session_handle in `ScreenCast.CreateSession`

Previously sessions were created via a `session_handle_token` in the
respective portal's `CreateSession` method. With the `SessionFactory` portal,
we can now create an independent Session and pass the object path
into `CreateSession` as `session_handle`. This allows us to re-use existing
sessions for multiple portals.

The existing API is updated to reflect that, in particular a session may
now have multiple types that can be checked invididually. For the
portal-specific API, the existing functions are deprecated and replaced
by one named after the portal, e.g. `xdp_session_start()` becomes
`xdp_remote_desktop_start()` or `xdp_screencast_start()`.

The `RemoteDesktop` and `ScreenCast` portals however are a bit special,
they're quite intertwined (and partially equivalent) and the session
created for one transparently becomes the other, depending primarily on
`SelectSources`. This leads to some duplication in the public API.

We want new applications to use the new API, so libportal transparently
implements support for older portals by returning a special `XdpSession`
that is unbound but gets created on DBus when calling that portal's
`CreateSession` later. This session object cannot be used for future
session-based portals that require `SessionPortal` to be present.

## Extra note

As for the (currently missing) implementation it's hopefully not too hard to imagine, the `XdpSession` just contains whatever private data the portals need and various checks for the session type prevent wrong access.

The `Location` API would likely need to be updated too for consistency, right now it doesn't allow for multiple sessions.

### DIFF
--- a/libportal/meson.build
+++ b/libportal/meson.build
@@ -19,6 +19,7 @@ headers = [
   'print.h',
   'remote.h',
   'screenshot.h',
+  'session.h',
   'spawn.h',
   'trash.h',
   'types.h',
--- a/libportal/remote.h
+++ b/libportal/remote.h
@@ -20,13 +20,23 @@
 #pragma once
 
 #include <libportal/types.h>
+#include <libportal/session.h>
 
 G_BEGIN_DECLS
 
-#define XDP_TYPE_SESSION (xdp_session_get_type ())
-
-XDP_PUBLIC
-G_DECLARE_FINAL_TYPE (XdpSession, xdp_session, XDP, SESSION, GObject)
+/**
+ * XdpSessionState:
+ * @XDP_SESSION_INITIAL: the session has not been started.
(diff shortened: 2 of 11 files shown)


whot:
Closing, this isn't going to happen and meanwhile the InputCapture portal can go on with subclassing the session, so that's good enough here.

PR #114, opened 2022-09-07
Support RemoteDesktop.ConnectToEIS

whot:
See https://github.com/flatpak/xdg-desktop-portal/pull/762

This adds support for connecting a remote desktop/screen cast session to an [EIS implementation](https://gitlab.freedesktop.org/libinput/libei/) so the application and the EIS implementation (i.e. the compositor) can talk directly without having to bounce DBus messages through the portal.

cc @jadah, @ofourdah

### DIFF
--- a/libportal/remote.c
+++ b/libportal/remote.c
@@ -898,6 +898,16 @@ xdp_session_open_pipewire_remote (XdpSession *session)
   return g_unix_fd_list_get (fd_list, fd_out, NULL);
 }
 
+static inline gboolean
+is_active_remote_desktop_session (XdpSession    *session,
+                                  XdpDeviceType  required_device)
+{
+  return XDP_IS_SESSION (session) &&
+         session->type == XDP_SESSION_REMOTE_DESKTOP &&
+         session->state == XDP_SESSION_ACTIVE &&
+         (session->devices & required_device) != 0;
+}
+
 /**
  * xdp_session_pointer_motion:
  * @session: a [class@Session]
@@ -916,10 +926,7 @@ xdp_session_pointer_motion (XdpSession *session,
 {
   GVariantBuilder options;
 
-  g_return_if_fail (XDP_IS_SESSION (session) &&
-                    session->type == XDP_SESSION_REMOTE_DESKTOP &&
-                    session->state == XDP_SESSION_ACTIVE &&
-                    ((session->devices & XDP_DEVICE_POINTER) != 0));
+  g_return_if_fail (is_active_remote_desktop_session (session, XDP_DEVICE_POINTER));
 
   g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
   g_dbus_connection_call (session->portal->bus,
@@ -952,10 +959,7 @@ xdp_session_pointer_position (XdpSession *session,
 {
   GVariantBuilder options;
 
-  g_return_if_fail (XDP_IS_SESSION (session) &&
-                    session->type == XDP_SESSION_REMOTE_DESKTOP &&
-                    session->state == XDP_SESSION_ACTIVE &&
-                    ((session->devices & XDP_DEVICE_POINTER) != 0));
+  g_return_if_fail (is_active_remote_desktop_session (session, XDP_DEVICE_POINTER));
 
   g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
   g_dbus_connection_call (session->portal->bus,
@@ -985,10 +989,7 @@ xdp_session_pointer_button (XdpSession *session,
 {
   GVariantBuilder options;
 
-  g_return_if_fail (XDP_IS_SESSION (session) &&
-                    session->type == XDP_SESSION_REMOTE_DESKTOP &&
-                    session->state == XDP_SESSION_ACTIVE &&
-                    ((session->devices & XDP_DEVICE_POINTER) != 0));
+  g_return_if_fail (is_active_remote_desktop_session (session, XDP_DEVICE_POINTER));
 
   g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
   g_dbus_connection_call (session->portal->bus,
@@ -1022,10 +1023,7 @@ xdp_session_pointer_axis (XdpSession *session,
 {
   GVariantBuilder options;
 
-  g_return_if_fail (XDP_IS_SESSION (session) &&
-                    session->type == XDP_SESSION_REMOTE_DESKTOP &&
-                    session->state == XDP_SESSION_ACTIVE &&
-                    ((session->devices & XDP_DEVICE_POINTER) != 0));
+  g_return_if_fail (is_active_remote_desktop_session (session, XDP_DEVICE_POINTER));
 
   g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
   g_variant_builder_add (&options, "{sv}", "finish", g_variant_new_boolean (finish));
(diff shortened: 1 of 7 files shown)


ofourdan:
You may want to cherry-pick https://github.com/ofourdan/libportal/commit/9689ea23 as well

whot:
done, thanks!

whot:
This is now merged into xdg-desktop-portal (https://github.com/flatpak/xdg-desktop-portal/pull/762) so let's mark this PR as ready.

libportal/remote.c
```diff
@@ -377,7 +418,7 @@ get_screencast_interface_version_returned (GObject *object,
   g_variant_get_child (ret, 0, "v", &version_variant);
   call->portal->screencast_interface_version = g_variant_get_uint32 (version_variant);
 
-  create_session (call);
+  get_remote_desktop_interface_version(call);
```
GeorgesStavracas, line 484: Nit: missing space before `(`

libportal/remote.c
```diff
@@ -516,7 +558,11 @@ xdp_portal_create_remote_desktop_session (XdpPortal *portal,
   call->multiple = (flags & XDP_REMOTE_DESKTOP_FLAG_MULTIPLE) != 0;
   call->task = g_task_new (portal, cancellable, callback, data);
 
-  create_session (call);
+  if (portal->screencast_interface_version == 0 ||
+      portal->remote_desktop_interface_version == 0)
+    get_screencast_interface_version (call);
```
GeorgesStavracas, line 626: `get_screencast_interface_version()` does not check if the screencast interface version was already fetched. If the operation is cancelled between `get_screencast_interface_version()` and `get_remote_desktop_interface_version()`, we'll end up with `portal->screencast_interface_version = 6` and ``portal->remote_desktop_interface_version = 0`, and this will fetch the interface version again unnecessarily.

It's not a big deal, but I'd rather make this code tighter. I think the simplest approach is adding a new function as follows:

```c
static void
get_portal_interface_versions (CreateCall *call)
{
  g_assert (call != NULL);
  g_assert (call->portal != NULL);

  if (call->portal->screencast_interface_version == 0)
    get_screencast_interface_version (call);
  else if (call->portal->remote_desktop_interface_version == 0)
    get_remote_desktop_interface_version (call);
  else
    create_session (call);
}
```

libportal/remote.c
```diff
@@ -892,6 +892,16 @@ xdp_session_open_pipewire_remote (XdpSession *session)
   return g_unix_fd_list_get (fd_list, fd_out, NULL);
 }
 
+static gboolean
+is_active_remotedesktop_session (XdpSession    *session,
```
GeorgesStavracas, line 959: Nit: `remote_desktop`

libportal/remote.c
```diff
@@ -892,6 +892,16 @@ xdp_session_open_pipewire_remote (XdpSession *session)
   return g_unix_fd_list_get (fd_list, fd_out, NULL);
 }
 
+static gboolean
```
GeorgesStavracas, line 958: `static inline gboolean`

libportal/remote.c
```diff
@@ -41,6 +42,68 @@ typedef struct {
...
+ *
+ * This call must be issued before xdp_session_start(). If successful, all input
+ * event emulation must be handled via the EIS connection and calls to
+ * xdp_session_pointer_motion() etc. are silently ignored.
+ *
+ * Returns: the file descriptor to the EIS implementation or a negative errno
+ * on failure.
+ */
+int
+xdp_session_connect_to_eis(XdpSession *session)
```
GeorgesStavracas, line 61: Nit: missing space before `(`

libportal/remote.c
```diff
@@ -41,6 +42,68 @@ typedef struct {
...
+ * This fd can be passed into ei_setup_backend_fd(). See the libei
+ * documentation for details.
+ *
+ * This call must be issued before xdp_session_start(). If successful, all input
+ * event emulation must be handled via the EIS connection and calls to
+ * xdp_session_pointer_motion() etc. are silently ignored.
+ *
+ * Returns: the file descriptor to the EIS implementation or a negative errno
+ * on failure.
+ */
```
GeorgesStavracas, line 59: Can you move this whole function to above `xdp_session_pointer_motion ()` ?

libportal/remote.c
```diff
@@ -41,6 +42,68 @@ typedef struct {
...
+ * Connect this XdpRemoteDesktopSession to an EIS implementation and return the fd.
+ * This fd can be passed into ei_setup_backend_fd(). See the libei
+ * documentation for details.
+ *
+ * This call must be issued before xdp_session_start(). If successful, all input
+ * event emulation must be handled via the EIS connection and calls to
+ * xdp_session_pointer_motion() etc. are silently ignored.
+ *
+ * Returns: the file descriptor to the EIS implementation or a negative errno
+ * on failure.
```
GeorgesStavracas, line 58: Hm... returning `int`s is not a pattern we use in libportal. I think it would be better to return a `gboolean`, and add a outgoing GError argument to this function. Something like:

```c
gboolean
xdp_session_connect_to_eis (XdpSession  *session
                            GError     **error)
```

GeorgesStavracas (changes requested):

libportal/remote.c
whot: done! in this case we do still need to return the `fd`, so I've changed this now to
```
int
xdp_session_connect_to_eis (XdpSession  *session
                            GError     **error)
```
and the return value is just `-1` in case of error. which roughly matches the existing
```
int xdp_session_open_pipewire_remote (XdpSession *session)
```
but that one doesn't (need to?) return a `GError`. I've simplified the error values a bit though so the caller either get's `NOT_SUPPORTED` or just `INVALID_VALUE`.

libportal/remote.c
whot: done, thanks. And all the other style changes are done as requested too

libportal/remote.c
```diff
@@ -905,9 +906,74 @@ is_active_remote_desktop_session (XdpSession    *session,
...
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED, "Not supported by the portal interface");
+  else if (session->type != XDP_SESSION_REMOTE_DESKTOP)
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_ARGUMENT, "Session is not a Remote Desktop session");
+  else if (xdp_session_get_session_state (session) != XDP_SESSION_ACTIVE)
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_ARGUMENT, "Session has not been started");
+  else if (session->uses_eis)
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_ARGUMENT, "Session is already connected to EIS");
+
+  if (error != NULL && *error != NULL)
+      return -1;
```
GeorgesStavracas, line 948: Sorry to be such a pain. I think these checks misbehave when someone calls `xdp_session_connect_to_eis (session, NULL);`. If NULL is passed, then this code will proceed to call `g_dbus_connection_call_with_unix_fd_list_sync()` regardless of having an error.

I think you have to either make all checks be something like this:

```c
if (portal->foo < bar)
  {
    g_set_error (error, <some error>);
    return -1;
  }
```

Or use an auxiliary variable:

```c
g_autoptr(GError) local_error = NULL;

if (portal->foo < bar)
  g_set_error (&local_error, <some error>);
else if (portal->frob < nicate)
  g_set_error (&local_error, <some error>);

if (local_error != NULL)
  {
    g_propagate_error (error, g_steal_pointer (&local_error);
    return -1;
  }
```

I don't have any strong preference, I'm slightly tending towards the first example as it would avoid allocating and deallocating a GError, but either should be fine.

(Also style nit: all `g_set_error`s and the `return -1;` are one indentation level off)

whot: oh, right, good catch. I tried to be too smart for my own good (had the `{...return -1; }` first and then tried to polish it - which also explains the indentation.

Done in the first style now, thanks for catching this. Unfortunately the python tests didn't because the PyGI bindings always pass a GError.

GeorgesStavracas (approved):

PR #115, opened 2022-10-12
Background status

GeorgesStavracas:
This is the libportal counterpart to https://github.com/flatpak/xdg-desktop-portal/pull/901

Made with love

### DIFF
--- a/libportal/background.c
+++ b/libportal/background.c
@@ -20,8 +20,115 @@
 
 #include "config.h"
 
+#include "session-private.h"
 #include "background.h"
 #include "portal-private.h"
+
+typedef struct {
+  XdpPortal *portal;
+  GTask *task;
+  char *status_message;
+} SetStatusCall;
+
+static void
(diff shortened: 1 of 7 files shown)


libportal/background.c
```diff
@@ -20,9 +20,115 @@
...
+  GTask *task;
+  char *status_message;
+} SetStatusCall;
+
+static void
+set_status_call_free (SetStatusCall *call)
+{
+  g_clear_pointer (&call->status_message, g_free);
+  g_clear_object (&call->portal);
+  g_clear_object (&call->task);
```
jadahl, line 38: Should `g_free (call);` too shouldn't it?

GeorgesStavracas: Correct, my bad

jadahl (approved):
lgtm now with the leak fixed.

GeorgesStavracas:
Thanks!

matthiasclasen:
This looks fine to me. Thanks for doing that work!

Issue #118, opened 2023-01-27
Report error instead of aborting when g_bus_get_sync fails


marler8997:
In `xdp_portal_init`, the `g_bus_get_sync` can fail if a system has not properly set up DBUS.  Instead of aborting, it would be nice if there was a way for the application to detect/handle this error, for example, reporting to the user that DBUS may not have been configured/setup properly.

It seems the reason for calling "abort" here could be because the Glib "init" hook doesn't provide a way to report an error?  Maybe we'd have to add a function to check for errors on an XdpPortal object after it has been initialized?

TingPing:
> the Glib "init" hook doesn't provide a way to report an error?

It does have a way: [`GInitable`](https://developer-old.gnome.org/gio/unstable/GInitable.html).

My assumption is just that it was a shortcut.

*All* of libportal is going to fail without DBus. So every API needs to handle a no-dbus situation and error. This would be a tedious thing to do.

That said I really don't like the `abort()` here. It is bad form for any library.

TingPing:
I guess we don't really need to make every API handle the situation.

Just implement `GInitable`, document that it is the responsibility of applications to check `g_initable_init()` and not use the object if it fails.

PR #119, opened 2023-01-27
portal: use GInitable to enable init error propagation

marler8997:
Fix issue #118

Adds `xdp_portal_initable_new` as an alterative to `xdp_portal_new` that can return a GError.

I used https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gcharsetconverter.c as a reference to see how to implement the `GInitable` interface.

### DIFF
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -69,8 +69,11 @@ enum {
 };
 
 static guint signals[LAST_SIGNAL];
+static void xdp_portal_initable_iface_init (GInitableIface  *iface);
 
-G_DEFINE_TYPE (XdpPortal, xdp_portal, G_TYPE_OBJECT)
+G_DEFINE_TYPE_WITH_CODE (XdpPortal, xdp_portal, G_TYPE_OBJECT,
+                         G_IMPLEMENT_INTERFACE (G_TYPE_INITABLE,
+                                                xdp_portal_initable_iface_init))
 
 static void
 xdp_portal_finalize (GObject *object)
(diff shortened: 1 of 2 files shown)


libportal/portal.c
```diff
@@ -315,6 +326,33 @@ xdp_portal_init (XdpPortal *portal)
...
+{
+}
+
+/**
+ * xdp_portal_initable_new:
+ * @error: A GError location to store the error occurring, or NULL to ignore.
+ *
+ * Creates a new [class@Portal] object.
+ *
+ * Returns: a newly created [class@Portal] object
```
TingPing, line 350:
```suggestion
 * Returns: a newly created [class@Portal] object or NULL on error
```

libportal/portal.c
```diff
@@ -305,6 +312,10 @@ xdp_portal_init (XdpPortal *portal)
 
   if (error)
     {
+      if (out_error) {
+          *out_error = error;
+          return FALSE;
+      }
       g_critical ("Failed to create XdpPortal instance: %s\n", error->message);
       abort ();
```
TingPing, line 320: This branch shouldn't `abort()`, it should always return `FALSE`.

libportal/portal.c
```diff
@@ -305,6 +312,10 @@ xdp_portal_init (XdpPortal *portal)
 
   if (error)
     {
+      if (out_error) {
+          *out_error = error;
```
TingPing, line 316:
```suggestion
          *out_error = g_steal_pointer (&error);
```

libportal/portal.c
```diff
@@ -327,7 +365,7 @@ xdp_portal_init (XdpPortal *portal)
 XdpPortal *
 xdp_portal_new (void)
 {
-  return g_object_new (XDP_TYPE_PORTAL, NULL);
+  return xdp_portal_initable_new(NULL);
```
TingPing, line 368: This is an API break. It could never return `NULL` before.

I think you should move the `abort()` here. And document above:

```
If D-Bus is unavailable this API will abort. We recommend using `xdp_portal_initable_new()` to safely handle this failure.

Comment (TingPing) [reply libportal/portal.c:368]: Also move the `g_critical()` here.

Comment (marler8997) [reply libportal/portal.c]: Ok.  With this change we also no longer need a local `error` variable inside initable_new so it gets a little simpler. Code has been updated.  Thanks for the fixes and timely review.

File: libportal/portal.c [hunk 5]
```diff
@@ -290,44 +293,78 @@ create_bus_from_address (const char *address,
...
+ * We recommend using xdp_portal_initable_new() to safely handle this failure.
+ *
  * Returns: a newly created [class@Portal] object
  */
 XdpPortal *
 xdp_portal_new (void)
 {
-  return g_object_new (XDP_TYPE_PORTAL, NULL);
+  g_autoptr(GError) error = NULL;
+  XdpPortal *portal = xdp_portal_initable_new(&error);
```
TingPing, line 360:
```suggestion
  XdpPortal *portal = xdp_portal_initable_new (&error);
```

PR #122, opened 2023-05-18
CI: use the github action's set-safe-directory instead of a manual run

whot:
Now that we have checkout@v3, we can use that argument instead of manually calling out to git.

Fixes 2e1eadea760cbcef5ce388ed81421388e989489c

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,5 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Work around git safe directory check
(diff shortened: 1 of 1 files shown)


whot:
Ok, apparently this doesn't work as documented. First attempt specifying the path [failed](https://github.com/flatpak/libportal/actions/runs/5010899199/jobs/8981187706?pr=122) and the re-run without arguments (relying on the default which is supposed to be `true`) [failed](https://github.com/flatpak/libportal/actions/runs/5011020625/jobs/8981421883?pr=122) too.

So... closing this, ETIME for debugging this..

PR #128, opened 2023-10-09
remote: Add API to allow for persistence on RemoteDesktop sessions

### LINKED ISSUE #127: No API to create persistent RemoteDesktop sessions
Required for https://github.com/input-leap/input-leap/issues/1699 - if the screensaver kicks in the RemoteDesktop Session is closed. For InputLeap we want to reconnect when it's available again, but that requires restoring the session.

`xdp_portal_create_remote_desktop_session()` doesn't allow the persist mode and restore token like `xdp_portal_create_screencast_session()` does, we'll need some new API here.

whot:
Same as the ScreenCast API

Fixes #127

cc @jadahl

Marking as Draft until I've sold off the right to rename this to a more sensible function name to the highest bidder. Any suggestions?

### DIFF
--- a/libportal/remote.c
+++ b/libportal/remote.c
@@ -237,6 +237,12 @@ select_devices (CreateCall *call)
   g_variant_builder_init (&options, G_VARIANT_TYPE_VARDICT);
   g_variant_builder_add (&options, "{sv}", "handle_token", g_variant_new_string (token));
   g_variant_builder_add (&options, "{sv}", "types", g_variant_new_uint32 (call->devices));
+  if (call->portal->remote_desktop_interface_version >= 2)
+    {
+      g_variant_builder_add (&options, "{sv}", "persist_mode", g_variant_new_uint32 (call->persist_mode));
+      if (call->restore_token)
+        g_variant_builder_add (&options, "{sv}", "restore_token", g_variant_new_string (call->restore_token));
+    }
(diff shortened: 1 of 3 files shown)


GeorgesStavracas:
libportal is not API nor ABI stable, so I'd be okay with breaking API for the next stable release.

For now, I guess a `_full` variant of this function should be enough

jadahl:
> libportal is not API nor ABI stable, so I'd be okay with breaking API for the next stable release.

Has that been the case always? If that's the case, how can libportal be suitable for any application that intends to be both distributable via Flatpak and traditional distributions?

GeorgesStavracas:
> Has that been the case always?

That's how I've been approaching libportal development since at least 2021 :slightly_smiling_face: 

> If that's the case, how can libportal be suitable for any application that intends to be both distributable via Flatpak and traditional distributions?

Personally, I consider libportal suitable for Flatpak applications only... somewhat like a Rust / cargo dependency, you have to pinpoint the exact version you want in the manifest.

whot:
Renamed to `_full` and tests updated and one extra new test added, should be ready.

GeorgesStavracas (approved):

jadahl:
> That's how I've been approaching libportal development since at least 2021 🙂

That is not what I would have assumed. Anyway, filed https://github.com/flatpak/libportal/issues/129.

PR #130, opened 2023-10-27
background: annotate commandline as nullable

v1993:
Ran into this while working on a Vala project. Code contains an explicit null check on it, so this seems to be an intended possibility.


GeorgesStavracas (approved):

GeorgesStavracas:
Code LGTM, but please change the commit title prefix to `background: `

v1993:
Done.

GeorgesStavracas:
Thank you

PR #131, opened 2023-10-29
Update README for building portal tests

hfiguiere:


hfiguiere:
This doesn't address how to launch the gtk4 test.

PR #135, opened 2023-12-25
Support xdg-foreign with Qt 6

### LINKED ISSUE #126: Qt 6.5+ does support xdg-foreign
The comment says `TODO: QtWayland doesn't support xdg-foreign protocol yet` but that's not quite true (moreover, QTBUG-76983 is closed as a duplicate while QTBUG-73801 is about importing a parent, not exporting). Qt 6.5 supports xdg-foreign, it's just there's no public API yet (although I'm not sure they will ever make it public, that's not something regular applications are supposed to use I guess). It could be accessed in such way:
```
	if (const auto services = dynamic_cast<QGenericUnixServices*>(
			QGuiApplicationPrivate::platformIntegration()->services())) {
		handle = g_strdup(services->portalWindowIdentifier(w).toUtf8().constData());
	}
```
The API prefixes the handle with x11:/wayland: doing actually all the _xdp_parent_export_qt's job.
The following includes are needed:
```
#include <qpa/qplatformintegration.h>
#include <private/qguiapplication_p.h>
#include <private/qgenericunixservices_p.h>
```
And a `#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)` check perhaps.

Another way is to support xdg-foreign manually with libwayland-client + a custom wl_registry instance using wl_display from `qApp->nativeInterface<QNativeInterface::QWaylandApplication>()->display()` (6.5+) or `reinterpret_cast<wl_display*>(QGuiApplication::platformNativeInterface()->nativeResourceForIntegration(QByteArray("wl_display")))` (5.4.0+ or if replacing "wl_display" with "display" then 5.0.0+). QGuiApplication::platformNativeInterface needs a `#include <qpa/qplatformnativeinterface.h>`.

ilya-fedin:
Fixes #126

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          dnf install -y meson gcc gobject-introspection-devel gtk3-devel gtk4-devel gi-docgen vala git python3-pytest python3-dbusmock qt5-qtbase-devel qt5-qtx11extras-devel qt6-qtbase-devel
+          dnf install -y meson gcc gobject-introspection-devel gtk3-devel gtk4-devel gi-docgen vala git python3-pytest python3-dbusmock qt5-qtbase-devel qt5-qtx11extras-devel qt6-qtbase-devel qt6-qtbase-private-devel
       - name: Check out libportal
         uses: actions/checkout@v3
       - name: Configure libportal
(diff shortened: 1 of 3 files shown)


Conan-Kudo (approved):

Conan-Kudo:
The logic seems reasonable to me from a cursory look.

Conan-Kudo:
FYI @grulja

GeorgesStavracas:
Very much appreciated, thank you

PR #138, opened 2024-01-11
docs: Add favicons

A6GibKm:
See https://gitlab.gnome.org/GNOME/gi-docgen/-/merge_requests/201. Generated by
https://gitlab.gnome.org/World/design/emblem/-/merge_requests/40.

The icon used was portal-symbolic from the devkit, using colors #26A269 to #33D17A were taken from the source of the SVG.


A6GibKm:
image

A6GibKm:
Note that it does not work locally, unless one runs it with 

```
python3 -m http.server
```

GeorgesStavracas (approved):

A6GibKm:
@GeorgesStavracas Could you please trigger a rebuild of the docs? I don't think it happens automatically.

PR #139, opened 2024-01-12
ci: Add dedicated job for documentation pages

GeorgesStavracas:
The Fedora jobs build and deploy documentation twice, and that may be a problem.

Use a dedicated workflow with proper a deploy job. This uses the Github Pages deploy action.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,61 @@
+name: Documentation
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
+
+    steps:
+      - name: Install dependencies
+        run: |
+          dnf install -y \
+            gcc \
+            gi-docgen \
+            git \
+            gobject-introspection-devel \
+            meson \
+            python3-pytest \
+            python3-dbusmock
+
+      - name: Check out libportal
+        uses: actions/checkout@v4
+
+      - name: Configure libportal
+        run: meson setup --prefix=/usr builddir -Dtests=false -Dvapi=false
+
+      - name: Build libportal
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
The `Source` setting in [`Settings` > `Pages`](https://github.com/flatpak/libportal/settings/pages) must be changed to "GitHub Actions" before merging this

.github/workflows/pages.yml
```diff
@@ -0,0 +1,68 @@
...
+            qt5-qtx11extras-devel \
+            qt6-qtbase-devel \
+            qt6-qtbase-private-devel \
+            vala
+
+      - name: Check out libportal
+        uses: actions/checkout@v3
+
+      - name: Configure libportal
+        run: meson setup --prefix=/usr builddir -Dbackend-gtk3=enabled -Dbackend-gtk4=enabled -Dbackend-qt5=enabled -Dbackend-qt6=enabled
```
TingPing, line 43: I think you can disable all backends as they wouldn't affect the docs.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Build libportal
+        run: ninja -C builddir
```
TingPing, line 46:
```suggestion
        run: ninja -C builddir doc/libportal-1
```

Only build what is needed.

.github/workflows/pages.yml
TingPing, line 43:
```suggestion
          dnf install -y \
            gcc \
            gi-docgen \
            git \
            gobject-introspection-devel \
            meson 

      - name: Check out libportal
        uses: actions/checkout@v3

      - name: Configure libportal
        run: meson setup --prefix=/usr builddir
```

TingPing, line 43:
```suggestion
        run: meson setup --prefix=/usr builddir -Dtests=false -Dvapi=false
```

GeorgesStavracas:
I'll make an executive decision to merge this, as I've applied all comments from @TingPing and things seem to be working

PR #142, opened 2024-02-09
portal-test/gtk3: Allow to use the gtk3 portal test on any screen

hfiguiere:
- Put the whole UI in a scrolled window to avoid overflowing


hfiguiere:
It doesn't fit on my screen otherwise.

GeorgesStavracas (approved):

PR #143, opened 2024-02-09
settings: Added support for the settings portal

hfiguiere:
Also added tests.

~~Currently on top of #142~~ merged


libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+    g_debug ("DBus error");
+    return NULL;
+  }
+
+  g_variant_get (ret, "(v)", &inner);
+
+  return g_steal_pointer (&inner);
+}
+
+XdpSettings * _xdp_settings_new (XdpPortal *portal)
```
TingPing, line 173:
```suggestion
XdpSettings *
_xdp_settings_new (XdpPortal *portal)
```

libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+ * xdp_settings_read:
+ * @settings: the [class@Settings] object.
+ * @namespace: the namespace of the value.
+ * @key: the key of the value.
+ *
+ * Read a setting value within @namespace, with @key.
+ *
+ * Returns: (nullable) (transfer full): the value, or %NULL if not found.
+ */
+XDP_PUBLIC
```
TingPing, line 146: `XDP_PUBLIC` goes in the header.

libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+GVariant *xdp_settings_read (XdpSettings *settings, const char *namespace, const char *key)
```
TingPing, line 147:
```suggestion
GVariant *
xdp_settings_read (XdpSettings *settings, const char *namespace, const char *key)
```

libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+
+  g_variant_get_child (parameters, 0, "&s", &namespace);
+  g_variant_get_child (parameters, 1, "&s", &key);
+  g_variant_get_child (parameters, 2, "v", &value);
+
+  g_signal_emit (settings, signals[CHANGED], 0, namespace, key, value);
+}
+
+/**
+ * xdp_settings_read:
```
TingPing, line 137: I'd love to see GSettings style wrappers around this:

`char *xdp_settings_read_string()`
`int xdp_settings_read_int()`

That can be a follow up though.

But in preparation for that this could be named `xdp_settings_read_value()` maybe?

Then we could have `xdp_settings_read()` take a GVariant format string like `g_settings_get()`.

libportal/portal.c
```diff
@@ -542,3 +548,15 @@ xdp_portal_running_under_sandbox (void)
...
 }
+
+/**
+ * xdp_portal_get_settings:
+ *
+ * This function return a new settings accessed through the portal.
+ *
+ * Returns: (transfer full): a [class@Settings] new settings object.
+ */
+XdpSettings *xdp_portal_get_settings (XdpPortal *portal)
```
TingPing, line 559:
```suggestion
XdpSettings *
xdp_portal_get_settings (XdpPortal *portal)
```

libportal/portal.c
```diff
@@ -542,3 +548,15 @@ xdp_portal_running_under_sandbox (void)
 {
   return xdp_portal_running_under_flatpak () || xdp_portal_running_under_snap (NULL);
 }
+
+/**
+ * xdp_portal_get_settings:
+ *
+ * This function return a new settings accessed through the portal.
```
TingPing, line 555: Weird phrasing.

libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+
+  ret = g_dbus_connection_call_sync (settings->portal->bus,
+                                     PORTAL_BUS_NAME,
+                                     PORTAL_OBJECT_PATH,
+                                     SETTINGS_INTERFACE,
+                                     "ReadOne",
+                                     g_variant_new ("(ss)", namespace, key),
+                                     NULL,
+                                     G_DBUS_CALL_FLAGS_NONE,
+                                     -1,
```
TingPing, line 161: Lets set a timeout, I don't think it should be very long, like 5 seconds?

libportal/settings.c
```diff
@@ -0,0 +1,191 @@
...
+                                     NULL, &error);
```
TingPing, line 162: Let's expose `error` as an input so callers understand why it fails.

TingPing, line 162: Maybe also pass through `cancellable`.

libportal/settings.c
```diff
@@ -0,0 +1,194 @@
...
+  if (!ret) {
+    g_debug ("DBus error");
+    return NULL;
+  }
+
+  g_variant_get (ret, "(v)", &inner);
+
+  return g_steal_pointer (&inner);
+}
+
```
TingPing, line 174: We should add a function for `ReadAll` too.

libportal/portal.c
hfiguiere: rephrased

libportal/settings.c
hfiguiere: done

libportal/settings.c
hfiguiere: aswell

libportal/settings.c
hfiguiere: aswell

libportal/settings.c
hfiguiere: done

libportal/settings.c
hfiguiere: done. with tests

libportal/settings.c
```diff
@@ -0,0 +1,234 @@
...
+ * SPDX-License-Identifier: LGPL-3.0-only
+ *
+ * Authors:
+ *    Hubert Figuière <hub@figuiere.net>
+ */
+
+#include "config.h"
+
+#include <libportal/portal.h>
+#include <libportal/settings.h>
```
TingPing, line 26:
```suggestion
#include "portal.h"
#include "settings.h"
```

libportal/settings.c
```diff
@@ -0,0 +1,234 @@
...
+ * xdp_settings_read_value:
+ * @settings: the [class@Settings] object.
+ * @namespace: the namespace of the value.
+ * @key: the key of the value.
+ * @cancellable: a GCancellable or NULL.
+ * @error: return location for error or NULL.
+ *
+ * Read a setting value within @namespace, with @key.
+ *
+ * Returns: (nullable) (transfer full): the value, or %NULL if not
```
TingPing, line 146:
```suggestion
 * Returns: (transfer full): the value, or %NULL if not
```

https://gi.readthedocs.io/en/latest/annotations/giannotations.html

> note that return and out values which are only NULL when throwing an error should not be annotated as (nullable)

Same for ReadAll.

libportal/settings.c
```diff
@@ -0,0 +1,234 @@
...
+                                     g_variant_new ("(^as)", namespaces),
+                                     NULL,
+                                     G_DBUS_CALL_FLAGS_NONE,
+                                     5000,
+                                     cancellable, error);
+  if (!ret)
+    {
+      g_debug ("DBus error");
+      return NULL;
+    }
```
TingPing, line 208:
```suggestion
  if (!ret)
    {
      g_debug ("DBus error");
      return NULL;
    }
```

This debug message isn't helpful. Either create an inner `GError` so we can always print a message or just remove it.

libportal/settings.c
TingPing: I think `xdp_settings_read_value` fine on its own but then `xdp_settings_read_all` is out of place. But `xdp_settings_read_all_values` is also kinda awkward? Thoughts?

libportal/settings.c
hfiguiere: removed.

libportal/settings.c
hfiguiere: done

libportal/settings.c
hfiguiere: done

libportal/settings.c
hfiguiere: I have added these API. I kept the GCancellable an GError.

libportal/settings.h
```diff
@@ -0,0 +1,53 @@
...
+
+XDP_PUBLIC
+GVariant *xdp_settings_read_value (XdpSettings *settings, const char *namespace, const char *key, GCancellable *cancellable, GError **error);
+
+XDP_PUBLIC
+void
+xdp_settings_read (XdpSettings *settings, const char *namespace,
+                   const gchar *key,
+                   GCancellable *cancellable, GError **error,
+                   const gchar *format, ...);
```
TingPing, line 42: Reformat to one line.

TingPing (approved):
Looks great

libportal/settings.h
hfiguiere: done

PR #147, opened 2024-03-08
notification: Implement version2 of the interface

jsparber:
Implement the new portal frontend in libportal https://github.com/flatpak/xdg-desktop-portal/pull/1298

### DIFF
--- a/libportal/notification.c
+++ b/libportal/notification.c
@@ -72,6 +72,69 @@ ensure_action_invoked_connection (XdpPortal *portal)
                                            NULL);
 }
 
+static int
+bytes_to_memfd (const gchar  *name,
+                GBytes       *bytes,
+                GError      **error)
+{
+  g_autofd int fd = -1;
+  gconstpointer bytes_data;
+  gpointer shm;
+  gsize bytes_len;
+
+  fd = memfd_create (name, MFD_ALLOW_SEALING);
+  if (fd == -1)
+    {
+      int saved_errno = errno;
+
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "memfd_create: %s", g_strerror (saved_errno));
+      return -1;
+    }
+
+  bytes_data = g_bytes_get_data (bytes, &bytes_len);
+
+  if (ftruncate (fd, bytes_len) == -1)
+    {
+      int saved_errno = errno;
+
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "ftruncate: %s", g_strerror (saved_errno));
+      return -1;
+    }
+
+  shm = mmap (NULL, bytes_len, PROT_WRITE, MAP_SHARED, fd, 0);
+  if (shm == MAP_FAILED)
+    {
+      int saved_errno = errno;
+
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "mmap: %s", g_strerror (saved_errno));
+      return -1;
+    }
+
+  memcpy (shm, bytes_data, bytes_len);
+
+  if (munmap (shm, bytes_len) == -1)
+    {
+      int saved_errno = errno;
+
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "munmap: %s", g_strerror (saved_errno));
+      return -1;
+    }
+
+  return g_steal_fd (&fd);
+}
+
 typedef struct {
   GUnixFDList *fd_list;
   GVariantBuilder *builder;
@@ -140,6 +203,189 @@ call_done_cb (GObject      *source,
 }
 
 static void
+splice_cb (GObject      *source_object,
+           GAsyncResult *res,
+           gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
+  g_autoptr(GError) error = NULL;
+  GOutputStream *stream_out = G_OUTPUT_STREAM (source_object);
+
+  if (!g_output_stream_splice_finish (G_OUTPUT_STREAM (source_object), res, &error))
+    {
+      g_task_return_error (task, g_steal_pointer (&error));
+      return;
+    }
+
+  if (G_IS_MEMORY_OUTPUT_STREAM (stream_out))
+    {
+      g_autoptr(GBytes) bytes = NULL;
+      g_autoptr(GVariant) media_out = NULL;
+
+      bytes = g_memory_output_stream_steal_as_bytes (G_MEMORY_OUTPUT_STREAM (stream_out));
+      media_out = g_variant_ref_sink (g_variant_new ("(sv)", "bytes",
+                                                     g_variant_new_from_bytes (G_VARIANT_TYPE_BYTESTRING, bytes, TRUE)));
+
+      g_task_set_task_data (task, g_steal_pointer (&media_out), (GDestroyNotify) g_variant_unref);
+      g_task_return_int (task, -1);
+    }
+  else if (G_IS_UNIX_OUTPUT_STREAM (stream_out))
+    {
+      int fd;
+      fd = g_unix_output_stream_get_fd (G_UNIX_OUTPUT_STREAM (stream_out));
+      lseek (fd, 0, SEEK_SET);
+      g_task_return_int (task, fd);
+    }
+  else
+    {
+      g_assert_not_reached ();
+    }
+}
+
+static void
+file_read_cb (GObject      *source_object,
+              GAsyncResult *res,
+              gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
+  GOutputStream *stream_out = G_OUTPUT_STREAM (g_task_get_task_data (task));
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GFileInputStream) stream_in = NULL;
+
+  stream_in = g_file_read_finish (G_FILE (source_object), res, &error);
+  if (!stream_in)
+    {
+      g_task_return_error (task, g_steal_pointer (&error));
+      return;
+    }
+
+  g_output_stream_splice_async (stream_out,
+                                G_INPUT_STREAM (stream_in),
+                                G_OUTPUT_STREAM_SPLICE_CLOSE_SOURCE | G_OUTPUT_STREAM_SPLICE_CLOSE_TARGET,
+                                g_task_get_priority (task),
+                                g_task_get_cancellable (task),
+                                splice_cb,
+                                g_object_ref (task));
+}
+
+static void
+parse_media (GVariant            *media,
+             uint                 version,
+             GCancellable        *cancellable,
+             GAsyncReadyCallback  callback,
+             gpointer             user_data)
+{
+  g_autoptr(GTask) task = NULL;
+  g_autoptr(GVariant) value = NULL;
+  const char *key;
+
+  task = g_task_new (NULL, cancellable, callback, user_data);
+  g_task_set_source_tag (task, parse_media);
+
+  if (!g_variant_is_of_type (media, G_VARIANT_TYPE("(sv)")))
+    {
+      g_task_return_boolean (task, TRUE);
+      return;
+    }
+
+  g_variant_get (media, "(&sv)", &key, &value);
+
+  if (strcmp (key, "bytes") == 0 && version > 1)
+    {
+      g_autoptr(GBytes) bytes = NULL;
+      g_autoptr(GError) error = NULL;
+      g_autoptr(GVariant) media_out = NULL;
+      g_autoptr(GOutputStream) stream_out = NULL;
+      g_autofd int fd = -1;
+
+      bytes = g_variant_get_data_as_bytes (value);
+
+      fd = bytes_to_memfd ("notification-media", bytes, &error);
+      if (fd == -1)
+        {
+          g_task_return_error (task, g_steal_pointer (&error));
+          return;
+        }
+
+      g_task_return_int (task, g_steal_fd (&fd));
+    }
+  else if (strcmp (key, "file") == 0)
+    {
+      g_autoptr(GFile) file = NULL;
+      g_autoptr(GOutputStream) stream_out = NULL;
+
+      if (version < 2)
+        {
+          stream_out = g_memory_output_stream_new_resizable ();
+        }
+      else
+        {
+          g_autofd int fd = -1;
+
+          fd = memfd_create ("notification-media", MFD_ALLOW_SEALING);
+          if (fd == -1)
+            {
+              int saved_errno = errno;
+
+              g_task_return_new_error (task,
+                                       G_IO_ERROR,
+                                       g_io_error_from_errno (saved_errno),
+                                       "memfd_create: %s", g_strerror (saved_errno));
+              return;
+            }
+
+          stream_out = g_unix_output_stream_new (g_steal_fd (&fd), FALSE);
+        }
+
+      g_task_set_task_data (task, g_steal_pointer (&stream_out), g_object_unref);
+
+      file = g_file_new_for_commandline_arg (g_variant_get_string (value, NULL));
+      g_file_read_async (file,
+                         g_task_get_priority (task),
+                         cancellable,
+                         file_read_cb,
+                         g_object_ref (task));
+    }
+  else
+    {
+      g_task_set_task_data (task, g_variant_ref (media), (GDestroyNotify) g_variant_unref);
+      g_task_return_int (task, -1);
+    }
+}
+
+static GVariant*
+parse_media_finish (GAsyncResult  *result,
+                    GUnixFDList   *fd_list,
+                    GError       **error)
+{
+  GTask *task = G_TASK (result);
+  g_autofd int fd = -1;
+
+  g_return_val_if_fail (g_task_get_source_tag (task) == parse_media, NULL);
+
+  fd = g_task_propagate_int (G_TASK (result), error);
+
+  if (*error)
+    return NULL;
+
+  if (fd != -1)
+    {
+      int fd_in;
+
+      fd_in = g_unix_fd_list_append (fd_list, fd, error);
+      if (fd_in == -1)
+          return NULL;
+
+      return g_variant_ref_sink (g_variant_new ("(sv)", "file-descriptor", g_variant_new_handle (fd_in)));
+    }
+  else
+    {
+      g_assert (g_task_get_task_data (task) != NULL);
+      return g_variant_ref (g_task_get_task_data (task));
+    }
+}
+
+static void
 parse_buttons_v1 (GVariantBuilder *builder,
                   GVariant        *value)
 {
@@ -171,6 +417,54 @@ parse_buttons_v1 (GVariantBuilder *builder,
   g_variant_builder_close (builder);
   g_variant_builder_close (builder);
   g_variant_builder_close (builder);
+}
+
+static void
+parse_icon_cb (GObject      *source_object,
+               GAsyncResult *res,
+               gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
+  ParserData *data = g_task_get_task_data (task);
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) media_out = NULL;
+
+  media_out = parse_media_finish (res, data->fd_list, &error);
+
+  if (!media_out)
+    {
+      g_task_return_error (task, g_steal_pointer (&error));
+      return;
+    }
+
+  g_variant_builder_add (data->builder, "{sv}", "icon", media_out);
+
+  if (parser_data_release (data))
+    g_task_return_boolean (task, TRUE);
+}
+
+static void
+parse_sound_cb (GObject      *source_object,
+                GAsyncResult *res,
+                gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
+  ParserData *data = g_task_get_task_data (task);
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) media_out = NULL;
+
+  media_out = parse_media_finish (res, data->fd_list, &error);
+
+  if (!media_out)
+    {
+      g_task_return_error (task, g_steal_pointer (&error));
+      return;
+    }
+
+  g_variant_builder_add (data->builder, "{sv}", "sound", media_out);
+
+  if (parser_data_release (data))
+    g_task_return_boolean (task, TRUE);
 }
 
 static void
@@ -226,7 +520,29 @@ parse_notification (GVariant            *notification,
             }
         }
 
-      g_variant_builder_add (data->builder, "{sv}", key, value);
+      if (strcmp (key, "icon") == 0)
+        {
+          parser_data_hold (data);
+          parse_media (value,
+                       version,
+                       cancellable,
+                       parse_icon_cb,
+                       g_object_ref (task));
+
+        }
+      else if (strcmp (key, "sound") == 0)
+        {
+          parser_data_hold (data);
+          parse_media (value,
+                       version,
+                       cancellable,
+                       parse_sound_cb,
+                       g_object_ref (task));
+        }
+      else
+        {
+          g_variant_builder_add (data->builder, "{sv}", key, value);
+        }
     }
 
   if (parser_data_release (data))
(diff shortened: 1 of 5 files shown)


libportal/notification.c
```diff
@@ -131,33 +222,43 @@ xdp_portal_add_notification (XdpPortal *portal,
...
-                          -1,
-                          cancellable,
-                          call_done_cb,
-                          call_done_data);
+  g_dbus_connection_call_with_unix_fd_list (portal->bus,
+                                            PORTAL_BUS_NAME,
+                                            PORTAL_OBJECT_PATH,
+                                            "org.freedesktop.portal.Notification",
+                                            "AddNotification",
+                                            g_variant_ref_sink (g_variant_builder_end (&builder)),
```
swick, line 254: The docs say that the `parameters` argument is owned by the caller, so this call won't unref the GVariant and it is leaked.

The docs also say "If the parameters GVariant is floating, it is consumed" and the `GVariantBuilder` docs say that `end` returns "A new, floating, GVariant".

So the best thing to do is this:

```suggestion
                                            g_variant_builder_end (&builder),
```

libportal/notification.c
```diff
@@ -19,9 +19,79 @@
...
+      return;
+    }
+
+  path = g_file_get_path (file);
+  fd = g_open (path, O_RDONLY | O_CLOEXEC);
+
+  if (*fd_list == NULL)
+    *fd_list = g_unix_fd_list_new_from_array (&fd, 1);
+  else
+    fd_in = g_unix_fd_list_append (*fd_list, fd, NULL);
```
swick, line 60: That seems like a really roundabout way of handling things. Why not just `g_unix_fd_list_new` in the caller and pass it through to here?

libportal/notification.c
```diff
@@ -131,33 +222,43 @@ xdp_portal_add_notification (XdpPortal *portal,
...
+  g_dbus_connection_call_with_unix_fd_list (portal->bus,
+                                            PORTAL_BUS_NAME,
+                                            PORTAL_OBJECT_PATH,
+                                            "org.freedesktop.portal.Notification",
+                                            "AddNotification",
+                                            g_variant_ref_sink (g_variant_builder_end (&builder)),
+                                            NULL,
+                                            G_DBUS_CALL_FLAGS_NONE,
+                                            -1,
+                                            fd_list,
```
swick, line 258: " The data is owned by the caller of the function". This will leak.

libportal/notification.c
```diff
@@ -19,9 +19,79 @@
...
+                      GUnixFDList     **fd_list)
+{
+  gsize i;
+
+  for (i = 0; i < g_variant_n_children (notification); i++)
+    {
+      const char *key;
+      g_autoptr(GVariant) value = NULL;
+
+      g_variant_get_child (notification, i, "{&sv}", &key, &value);
```
swick, line 77: Looks like a good place to use g_variant_iter_loop/g_variant_iter_next

libportal/notification.c
swick, line 60: > The file descriptor is duplicated using dup(). You keep your copy of the descriptor and the copy contained in list will be closed when list is finalized.

The fd is leaking. Use g_autofd for fd.

swick: You could steal this https://github.com/flatpak/xdg-desktop-portal/blob/main/src/xdp-utils.h#L178

swick: or bump the version but I'm not sure what the policy is here

libportal/notification.c
```diff
@@ -36,7 +36,7 @@ prepare_media (GVariantBuilder *builder,
   g_autoptr(GFile) file = NULL;
   const char *key;
   g_autofree char *path = NULL;
-  g_autofd int fd = -1;
+  int fd = -1;
```
swick: should be in the previous commit

swick:
Might want to put the autofd stuff into a new commit but otherwise this LGTM.

libportal/notification.c
```diff
@@ -180,6 +181,9 @@ call_done (GObject *source,
  * - default-action-target `v`: target parameter to send along when
  *     activating the default action.
  * - buttons `aa{sv}`: array of serialized buttons
+ * - desktop-file-id `s`: A desktop file id containing the .desktop suffix.
+ * - display-hint `as`: An array of display hints.
+ * - category `s`: A category for this notification.
```
TingPing, line 186: Could you include a link to valid hints and categories?

jsparber: done

libportal/notification.c
```diff
@@ -94,7 +167,8 @@ call_done (GObject *source,
  *
  * - title `s`: a user-visible string to display as title
  * - body `s`: a user-visible string to display as body
- * - icon `v`: a serialized icon (in the format produced by [method@Gio.Icon.serialize])
+ * - icon `v`: a serialized icon (in the format produced by [method@Gio.Icon.serialize]
```
davidedmundson, line 173: This needs to be a defined standard rather than explicitly requiring a library.

swick: The serialized icon is a requirement for v1 of the portal which is unfortunate. I guess we could take an arbitrary piece of memory for this API though and convert it to the serialized icon when only v1 is supported.

swick:
xdg-desktop-portal now depends on this PR

libportal/portal-private.h
```diff
@@ -71,3 +73,30 @@ const char * portal_get_bus_name (void);
...
+  if (fd >= 0)
+    {
+      errsv = errno;
+      if (close (fd) < 0)
+        g_assert (errno != EBADF);
+      errno = errsv;
+    }
+}
+
+#define portal_autofd __attribute__((cleanup(portal_close_fd)))
```
GeorgesStavracas, line 102: `portal_` → `xdp_`

libportal/notification.c
```diff
@@ -19,9 +19,82 @@
...
+#include <gio/gunixfdlist.h>
+#include <gio/gunixoutputstream.h>
+
 #include "notification.h"
 #include "portal-private.h"
 
+static void
+prepare_media (GVariantBuilder *builder,
+              GVariant         *serialized_media,
+              GUnixFDList      *fd_list)
```
GeorgesStavracas, line 36: Misaligned arguments

libportal/notification.c
```diff
@@ -169,13 +171,22 @@ call_done (GObject *source,
...
  *     will be activated when the user clicks on the notification
  * - default-action-target `v`: target parameter to send along when
  *     activating the default action.
  * - buttons `aa{sv}`: array of serialized buttons
  *
+ * The serialized sound consists of a `s` or `sv`:
+ * - default : Play the default sound for the notification.
+ * - silent : Don't ever play a sound for the notification.
+ * - file `s`: A path to a sound file.
+ * - bytes `ay`: An array of bytes.
```
GeorgesStavracas, line 190: Is this still supported?

libportal/notification.c
```diff
@@ -306,3 +306,44 @@ xdp_portal_remove_notification (XdpPortal  *portal,
...
+/**
+ * xdp_portal_get_supported_options:
+ * @portal: a [class@Portal]
+ *
+ * Returns: a vardict of supported options for properties that have options.
+ *
+ * Since: 0.7.2
+ */
+GVariant *
+xdp_portal_get_supported_options (XdpPortal *portal)
```
GeorgesStavracas, line 324: Hm, I'm not particularly fond of just exposing the raw GVariant... I think either libportal provides some "added value" by exposing a nice API around it (XdpNotificationOptions?), or we let the app check that using their favorite tools. This is a simple D-Bus call anyway.

(Also, `xdp_get_supported_notification_options` is probably a better name?)

GeorgesStavracas (changes requested):

libportal/portal-private.h
jsparber: i renamed it on purpose. What's the motivation to use `xdp_`?

libportal/notification.c
jsparber: @davidedmundson Libportal provides GIO-style API for the portals so it's fine to depend on GIO.

@swick not sure what you mean? do you suggest to change the API?

libportal/portal-private.h
jsparber: just realized everything in libportal is prefixed with xdp not portal.

libportal/notification.c
jsparber: > This is a simple D-Bus call anyway.

Isn't that the case for the entire portal? Anyhow I'm fine with adding `XdpNotificationOptions`

> (Also, xdp_get_supported_notification_options is probably a better name?)

yes, definitely.

libportal/notification.c
swick: I think the point I was trying to make is that we can now send an arbitrary blob of data for icons and sounds without having to serialize them. This API doesn't allow for that.

Let's not get hung up on this and try to get the existing changes here into shape to get it merged.

davidedmundson: ack, I didn't read the repo name properly. This is a non-issue.

libportal/notification.c
jsparber: As discussed on matrix let's not create to much abstraction over the interface.

jsparber:
I reworked this MR to keep backwards compatibility. I didn't fully test it yet, therefore i'm marking it as a draft.

jsparber:
Should be ready for review. I changed the MR to keep backwards compatibility and some tests to make sure it actually works.

libportal/portal-private.h
```diff
@@ -71,3 +73,30 @@ const char * portal_get_bus_name (void);
 #define FLATPAK_PORTAL_BUS_NAME "org.freedesktop.portal.Flatpak"
 #define FLATPAK_PORTAL_OBJECT_PATH "/org/freedesktop/portal/Flatpak"
 #define FLATPAK_PORTAL_INTERFACE "org.freedesktop.portal.Flatpak"
+
+static inline int
```
bilelmoussaoui, line 79: What is the difference with g_autofd? it would be nice if the commit explains why.

jsparber: moved the first two commits to a different MR since it's not relevant here anymore https://github.com/flatpak/libportal/pull/172

jsparber: added a comment about the reason why we can't use g_autofd to the other mr

libportal/notification.c
```diff
@@ -60,22 +70,155 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+
+  vardict = g_variant_get_child_value (ret, 0);
+  g_variant_iter_init (&iter, vardict);
 
-  call_done_data->callback (G_OBJECT (call_done_data->portal), result, call_done_data->data);
+  while (g_variant_iter_loop (&iter, "{&sv}", &key, &value))
+    {
+      if (strcmp (key, "version"))
+        portal->notification_interface_version = g_variant_get_uint32 (value);
+    }
```
swick: Looks like a convoluted way to write this:
```c
g_variant_get (ret, "(&a{sv})", &vardict);
g_variant_lookup (vardict, "version", "u", &portal->notification_interface_version);
```

libportal/notification.c
```diff
@@ -129,35 +272,30 @@ xdp_portal_add_notification (XdpPortal *portal,
...
-                          call_done_cb,
-                          call_done_data);
+  g_variant_builder_add (builder, "s", id);
+  /* This will be closed once the async parsing is done */
+  g_variant_builder_open (builder, G_VARIANT_TYPE ("a{sv}"));
+
+  g_task_set_source_tag (task, xdp_portal_add_notification);
+  g_task_set_task_data (task, call_data_new (iter, builder), call_data_free);
+
+  if (portal->notification_interface_version < 1)
```
swick, line 717: `== 0`?

libportal/notification.c
```diff
@@ -129,35 +272,30 @@ xdp_portal_add_notification (XdpPortal *portal,
...
+    get_notification_interface_properties (task);
+  else
+    parse_notification (task);
```
swick, line 720: You could change the `get_notification_interface_properties` to be async which can finish immediately in some situations and when the dbus call is done otherwise. This allows you to have a single flow:
```c
xdp_portal_add_notification()
{
  get_supported(..., get_supported_cb);
}

get_supported_cb()
{
  finish_get_supported();
  parse_notification();
}
```

libportal/notification.c
```diff
@@ -60,22 +70,155 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+  GVariant *value_outer = NULL;
+
+  while (g_variant_iter_next (data->iter, "{&sv}", &key, &value_outer))
+    {
+      g_autoptr(GVariant) value = value_outer;
+
+      /* In version 1 of the interface requests fail if there are unsupported properties */
+      if (portal->notification_interface_version < 2)
+        {
+          const char *supported_properties_v1[] = {
```
swick, line 486: Probably better to move this out of the loop.

libportal/notification.c
```diff
@@ -60,22 +70,155 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+  res = g_dbus_connection_call_with_unix_fd_list_finish (portal->bus, NULL, result, &error);
+
+  if (res)
+    g_task_return_boolean (task, TRUE);
+  else
+    g_task_return_error (task, g_steal_pointer (&error));
+}
+
+static void
+parse_notification (GTask *task)
```
swick, line 472: It's a bit unexpected that a function named `parse_notification` does a dbus call. Maybe split the actual "parsing" out of here.

libportal/notification.c
```diff
@@ -129,35 +272,30 @@ xdp_portal_add_notification (XdpPortal *portal,
...
-                          -1,
-                          cancellable,
-                          call_done_cb,
-                          call_done_data);
+  g_variant_builder_add (builder, "s", id);
+  /* This will be closed once the async parsing is done */
+  g_variant_builder_open (builder, G_VARIANT_TYPE ("a{sv}"));
+
+  g_task_set_source_tag (task, xdp_portal_add_notification);
+  g_task_set_task_data (task, call_data_new (iter, builder), call_data_free);
```
swick, line 715: Doesn't change the semantics but iter and builder are both owned so maybe use g_autoptr/g_steal_pointer.

libportal/notification.c
```diff
@@ -195,10 +598,15 @@ get_properties_cb (GObject      *source_object,
 
   while (g_variant_iter_loop (&iter, "{&sv}", &key, &value))
     {
-      if (strcmp (key, "version"))
+      if (strcmp (key, "version") == 0)
         portal->notification_interface_version = g_variant_get_uint32 (value);
+      else if (strcmp (key, "SupportedOptions") == 0)
+        portal->supported_notification_options = g_steal_pointer (&value);
```
swick, line 604: This will leak a previously set `portal->supported_notification_options`.

libportal/notification.c
```diff
@@ -148,12 +497,66 @@ parse_notification (GTask *task)
...
+          if (parse_icon_property (task, value))
+            return;
+        }
+      else if (strcmp (key, "sound") == 0)
+        {
+          /* This parse loop will be restarted in case of async loading */
+          if (parse_sound_property (task, value))
+            return;
+        }
+      else if (portal->notification_interface_version < 2 && strcmp (key, "buttons") == 0)
```
swick, line 512: `>=2`, no?

libportal/notification.c
```diff
@@ -148,12 +497,66 @@ parse_notification (GTask *task)
...
+        {
+          GVariant *button;
```
swick, line 514: Also mayve move to its own function.

libportal/notification.c
```diff
@@ -119,6 +191,283 @@ call_done_cb (GObject      *source,
...
+    {
+      g_variant_builder_add (data->builder, "{sv}", "icon", serialized_icon);
+
+      return FALSE;
+    }
+}
+
+static gboolean
+parse_sound_property (GTask            *task,
+                      GVariant         *serialized_sound)
```
swick, line 392: It's a bit unnecessary to pass the task all the way here when all you need it the builder and return an error.

jsparber: `parse_sound_property()` does async and It needs to be passed task to `file_read_cb()`

libportal/notification.c
jsparber: No, before version 2 we need to filter out properties version 1 doesn't support.

libportal/notification.c
jsparber: true, but this function is only called once. Expect if xdp doesn't have a version property. So maybe still good to free the previous value.

libportal/notification.c
jsparber: I prefer to not loop twice over the vardict, since we have two properties we need ultimately.

libportal/notification.c
jsparber: no strong opinion

libportal/notification.c
jsparber: i can rename the function to `handle_notification()`

libportal/notification.c
jsparber: I don't understand what you suggest

libportal/notification.c
swick: I think it would be more understandable if you first collect all files that you want to read, then do the reading, then do the "parsing".

jsparber: yes, that's what i had initially. But for v1 we need to convert `GFileIcon` to `GBytesIcon` which has the icon content inline. The I think the only two option for that is to parse the vardicts twice, the second time adding the bytes, or modify the `GVariant` inplace, which sounds a bad thing to do.

libportal/notification.c
swick: Vardicts are not lists but more like hashtables. Even then, I don't think the increased complexity justifies some potential performance. Micro-optimizations without measurements are just not a good idea.

jsparber: fine with me

libportal/notification.c
swick: The flows through the code are a bit messy and could be organized more linearly by implementing a bunch of tasks and then running one after the other, unconditionally:

1. get supported features
2. read files
3. add the notification

Then handle details in the tasks, for example "get supported features" would be something like this:

```c
static void
get_properties_cb (GObject      *source_object,
                   GAsyncResult *result,
                   gpointer      user_data)
{
  g_autoptr(GTask) task = G_TASK (user_data);
  XdpPortal *portal = g_task_get_source_object (task);
  g_autoptr(GError) error = NULL;
  g_autoptr(GVariant) ret = NULL;
  g_autoptr(GVariant) vardict = NULL;

  ret = g_dbus_connection_call_finish (G_DBUS_CONNECTION (source_object), result, &error);
  if (!ret)
    {
      g_task_return_error (task, g_steal_pointer (&error));
      return;
    }

  g_clear_pointer (&portal->supported_notification_options, g_variant_unref);

  g_variant_get (ret, "(&a{sv})", &vardict);
  if (!g_variant_lookup (vardict, "version", "u", &portal->notification_interface_version))
    portal->notification_interface_version = 1;
  if (!g_variant_lookup (vardict, "SupportedOptions", "v", &portal->supported_notification_options))
    {
      g_auto(GVariantBuilder) builder;
      
      g_variant_builder_init (&builder, G_VARIANT_TYPE_VARDICT);
      portal->supported_notification_options = g_variant_ref_sink (g_variant_builder_end (&builder));
    }

  g_task_return_boolean (task, TRUE);
}

void
get_supported_features (XdpPortal *portal,
                        GCancellable *cancellable,
                        GAsyncReadyCallback callback,
                        gpointer data)
{
  g_autoptr(GTask) task = NULL;

  task = g_task_new (portal, cancellable, callback, data);
  g_task_set_source_tag (task, get_supported_features);

  if (portal->notification_interface_version != 0)
    {
      g_task_return_boolean (task, TRUE);
      return;
    }

  g_dbus_connection_call (portal->bus,
                          PORTAL_BUS_NAME,
                          PORTAL_OBJECT_PATH,
                          "org.freedesktop.DBus.Properties",
                          "GetAll",
                          g_variant_new ("(s)", "org.freedesktop.portal.Notification"),
                          G_VARIANT_TYPE ("(a{sv})"),
                          G_DBUS_CALL_FLAGS_NONE,
                          -1,
                          cancellable,
                          get_properties_cb,
                          g_object_ref (task));
}

gboolean
get_supported_features_finish (XdpPortal     *portal,
                               GAsyncResult  *result,
                               GError       **error)
{
  g_autoptr(GVariant) res = NULL;
  g_return_val_if_fail (XDP_IS_PORTAL (portal), FALSE);
  g_return_val_if_fail (g_task_is_valid (result, portal), FALSE);
  g_return_val_if_fail (g_task_get_source_tag (G_TASK (result)) == get_supported_features, FALSE);

  return g_task_propagate_boolean (G_TASK (result), error);
}
```

So, in other words, embrace the GTask.

jsparber: ah ok, so add additional `GTask` instead of using a single one. i can do that.

libportal/notification.c
swick: Iterating over the GVariant once to figure out which files need to be read and then once more to build the new variant that is handed to dbus sounds completely fine to me.

libportal/notification.c
jsparber: hmm, you marked `get_supported_features()` as public. Is it just an oversight?

swick: Yes, I didn't even compile the code

libportal/notification.c
jsparber: Actually  `g_variant_lookup()` does a linear scan https://docs.gtk.org/glib/method.Variant.lookup.html

swick: Not ideal but there are literally 2 properties right now and it won't increase to a point where performance will become relevant.

libportal/notification.c
jsparber: reworked the code to use multiple `GTask` and to have a single code path.

libportal/notification.c
jsparber: I kept the parsing function still async, but without restarting the loop. The parse walks the vardict once and completes once the async loading of the icon/sound has finished.

libportal/notification.c
```diff
@@ -60,22 +73,312 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+
+    g_variant_iter_init (&iter_button, button);
+    while (g_variant_iter_loop (&iter_button, "{&sv}", &button_key, &button_value))
+    {
+      if (strcmp (button_key, "purpose") != 0)
+        g_variant_builder_add (builder, "{sv}", button_key, button_value);
+    }
+
+    g_variant_builder_close (builder);
+  }
```
GeorgesStavracas: This whole section is short of 1 indentation

libportal/notification.c
```diff
@@ -60,22 +73,312 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+    GVariantIter iter_button;
+    g_variant_builder_open (builder, G_VARIANT_TYPE ("a{sv}"));
+
+    g_variant_iter_init (&iter_button, button);
+    while (g_variant_iter_loop (&iter_button, "{&sv}", &button_key, &button_value))
+    {
+      if (strcmp (button_key, "purpose") != 0)
+        g_variant_builder_add (builder, "{sv}", button_key, button_value);
+    }
+
```
GeorgesStavracas, line 418: This one too

libportal/notification.c
```diff
@@ -60,22 +73,312 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+static gboolean
+parser_data_release (ParserData *data)
+{
+  data->parse_ref--;
+
+  if (data->parse_ref == 0)
+    g_variant_builder_close (data->builder);
+
+  return data->parse_ref == 0;
+}
```
GeorgesStavracas, line 178: The common terminology in GLib land is `ref` and `unref`. You could also use `gatomicrefcount` for this.

libportal/notification.c
```diff
@@ -72,6 +72,73 @@ ensure_action_invoked_connection (XdpPortal *portal)
...
+  gconstpointer bytes_data;
+  gpointer shm;
+  gsize bytes_len;
+
+  fd = memfd_create (name, MFD_ALLOW_SEALING);
+  if (fd == -1)
+    {
+      int saved_errno;
+
+      saved_errno = errno;
```
GeorgesStavracas, line 90: Nitpick:

```suggestion
      int saved_errno = errno;
```

throughout the file

GeorgesStavracas (changes requested):

libportal/notification.c
jsparber: I used hold/release since it extends the lifetime of parser similar to `g_application_hold/release()` But i'm also fine with changing it

jsparber: i renamed the `parse_ref` to make clear that this isn't actually freeing anything.

GeorgesStavracas (approved):

PR #150, opened 2024-05-22
remote: fix screencasting on wlroots-based compositors

MStarvik:
xdp_portal_create_screencast_session fails if the RemoteDesktop interface is unavailable, which is currently the case on all wlroots-based compositors. xdp_portal_create_screencast_session queries the version of the RemoteDesktop interface despite this interface not beings used in the screencast session. Likewise,
xdp_portal_create_remote_desktop_session queries the version of the ScreenCast interface despite this interface not being used in the remote desktop session. This commit removes the unnecessary version queries.

### DIFF
--- a/libportal/remote.c
+++ b/libportal/remote.c
@@ -424,7 +424,7 @@ get_screencast_interface_version_returned (GObject *object,
   g_variant_get_child (ret, 0, "v", &version_variant);
   call->portal->screencast_interface_version = g_variant_get_uint32 (version_variant);
 
-  get_remote_desktop_interface_version (call);
+  create_session (call);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
This removal will break the check at https://github.com/flatpak/libportal/blob/main/libportal/remote.c#L240-L245.

I agree it's not necessary to fetch the Remote Desktop version when calling ScreenCast functions, but this code cannot simply be removed. The Remote Desktop version fetching should only happen when calling remote desktop APIs instead.

MStarvik:
RemoteDesktop version fetching still happens when calling remote desktop APIs, and ScreenCast version fetching still happens when calling screencast APIs. Only the seemingly unnecessary version queries have been removed.

GeorgesStavracas (approved):

PR #161, opened 2024-09-03
Various CI updates

GeorgesStavracas:
See commits


.github/workflows/build.yml
```diff
@@ -24,8 +24,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt-get update
-          apt-get install -y libglib2.0 gettext dbus meson libgirepository1.0-dev libgtk-3-dev libgtk-4-dev valac python3-pip python3-dbusmock
-          pip3 install gi-docgen
```
smcv, line 41: I think the reason this was installed from pip was that Ubuntu 20.04 was too old to have a suitable version via apt.

smcv (approved):

.github/workflows/build.yml
GeorgesStavracas: Thanks, I've documented this in the commit message.

GeorgesStavracas:
Thanks Simon

PR #162, opened 2024-09-03
Unassorted fixes

GeorgesStavracas:
See commits

Depends on https://github.com/flatpak/libportal/pull/161 so that tests run in CI

### DIFF
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -987,7 +987,8 @@ set_pointer_barriers (Call *call)
 }
 
 static void
-gobject_ref_wrapper (gpointer data, gpointer user_data)
+gobject_ref_wrapper (gpointer data,
+                     gpointer user_data)
 {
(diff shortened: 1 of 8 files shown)


build-aux/org.gnome.PortalTest.Qt6.json
```diff
@@ -1,7 +1,7 @@
 {
     "app-id": "org.gnome.PortalTest.Qt6",
     "runtime": "org.kde.Platform",
-    "runtime-version": "6.5",
+    "runtime-version": "6.6",
```
hfiguiere, line 4: not sure if that can be done but current is 6.7. FYI.

GeorgesStavracas: This commit was part of #161 so I updated it there before merging. This PR is now up-to-date with that.

TingPing (approved):

PR #164, opened 2024-09-04
Fix an Input Capture test failure

GeorgesStavracas:
It's only reproducible on `meson dist` for some reason, `meson test` runs fine.


matthiasclasen:
Looks fine to me. Curious that it doesn't fail in meson test though

GeorgesStavracas:
Thanks!

PR #168, opened 2024-09-10
pyportaltests: work around potential session close timeout failures

### LINKED ISSUE #166: test_close_session_signal intermittently failing
I'm seeing intermittent test failures with 0.8.0 ([example full log](https://buildd.debian.org/status/fetch.php?pkg=libportal&arch=ppc64el&ver=0.8.0-1&stamp=1725366680&raw=0)) which still seem to be present in 0.8.1:

```
=================================== FAILURES ===================================
_________________ TestRemoteDesktop.test_close_session_signal __________________

self = <pyportaltest.test_remotedesktop.TestRemoteDesktop testMethod=test_close_session_signal>

    def test_close_session_signal(self):
        """
        Ensure that we get the GObject signal when our session is closed
        externally.
        """
        params = {"close-after-start": 500}
        setup = self.create_session(params=params)
        session = setup.session
    
        session_closed_signal_received = False
    
        def session_closed(session):
            nonlocal session_closed_signal_received
            session_closed_signal_received = True
            self.mainloop.quit()
    
        session.connect("closed", session_closed)
    
        self.mainloop.run()
    
>       assert session_closed_signal_received is True
E       assert False is True

pyportaltest/test_remotedesktop.py:599: AssertionError
```

I'm not sure I understand how this can happen: if `session_closed` didn't run, then how can we leave `self.mainloop.run` without having set `session_closed_signal_received`?

Is it possible that `self.mainloop` might sometimes be terminated by a leftover signal handler from a previous test, or something like that?

whot:
This may or may not fix #166 but ever optimistically we go for the former. 

cc @smcv, @GeorgesStavracas

### DIFF
--- a/tests/pyportaltest/test_remotedesktop.py
+++ b/tests/pyportaltest/test_remotedesktop.py
@@ -56,6 +56,7 @@ class TestRemoteDesktop(PortalTest):
         start_session=True,
         persist_mode=None,
         restore_token=None,
+        cb_closed_signal=None,
     ) -> SessionSetup:
         params = params or {}
         # To make the tests easier, load ScreenCast automatically if we have
@@ -119,6 +120,9 @@ class TestRemoteDesktop(PortalTest):
         session_handle = options["session_handle_token"]
 
         assert session.get_session_type() == Xdp.SessionType.REMOTE_DESKTOP
+
+        if cb_closed_signal:
+            session.connect("closed", cb_closed_signal)
 
         if outputs:
             # open the PW remote by default since we need it anyway for Start()
@@ -581,9 +585,6 @@ class TestRemoteDesktop(PortalTest):
         Ensure that we get the GObject signal when our session is closed
         externally.
         """
-        params = {"close-after-start": 500}
-        setup = self.create_session(params=params)
-        session = setup.session
 
         session_closed_signal_received = False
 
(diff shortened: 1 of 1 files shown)


smcv:
Thanks, I'll try applying this to the packages where I saw #166 and see what happens.

smcv:
This isn't solving the issue for me. When I run the test in a loop, I'm variously either seeing success, the same failure mode as before, or a segfault inside libportal.

Following up on the segfault is probably the best route, since fixing that might also fix #166.

smcv:
> Following up on the segfault is probably the best route

#169

whot:
Closing based on https://github.com/flatpak/libportal/pull/168#issuecomment-2343139942

whot:
@smcv I think I got it - this reliably passes on my machine now

whot:
Alright, this time it does work. So in the end it was a race condition between when the signal was sent out and when the test registered for the signal. The 500ms simply weren't enough.

Yesterday's patch had a bug - to make it nice python I dropped the result but that had the side-effect of Python cleaning up the `XdpSession` object (which was only exposed when I rebased on top of 0f8f627dd86b80f380dd6f0a02b11981a7c8a5e1). So keeping the `session` variable alive until the test fixes that particular issue now.

GeorgesStavracas (approved):

PR #178, opened 2024-11-28
Two more fixes for notification memory leaks

swick:


libportal/notification.c
```diff
@@ -580,7 +585,7 @@ parse_notification_cb (GObject      *source_object,
   XdpPortal *portal = g_task_get_source_object (task);
   g_autoptr(GError) error = NULL;
   g_autoptr(GUnixFDList) fd_list = NULL;
-  g_autoptr(GVariant) parameters = NULL;
+  GVariant *parameters = NULL;
```
GeorgesStavracas, line 588: Wouldn't it be more semantic to keep the `g_autoptr` here, and make `parse_notification_finish()` return `g_steal_pointer (&data->data);`?

jsparber:
Thansk for looking into this. Lgtm

GeorgesStavracas (approved):
One suggestion but overall it LGTM

libportal/notification.c
swick: Good point.

PR #181, opened 2025-01-02
Fix some Meson warnings by officially requiring 0.55.0

smcv:
* build: Officially require Meson 0.55.0
    
    In practice the GTK3 backend accidentally required 0.55.0 since 0.8.0,
    and nobody seems to have complained or even noticed.
    
    Fixes: f7f3acf5 "portal-test/gtk3: Don't provide a run target if we cannot run it"

* gtk4: Replace deprecated path() method with full_path()
    
    This raises a deprecation warning with newer Meson, and now that we
    officially require Meson 0.55.0, we can use it unconditionally.

* build: Never require a C++ compiler for the build architecture
    
    Newer Meson issues a warning if we don't specify whether we plan to
    compile code for the build architecture or not when adding a language.
    
    Like most projects, this one doesn't seem to ever compile for the build
    architecture, only for the host architecture, so when cross-compiling it
    will only need a host-architecture C++ compiler.
    
    For example, if we're cross-compiling for riscv64 on x86_64, we
    will need a riscv64 cross-compiler, but we don't need an x86_64 native
    compiler. Instruct Meson not to look for one.

---

Alternatively we could continue to only require 0.49.0 with some simple changes to the gtk3 and gtk4 backends, but Meson ≥ 0.55.0 is available since Debian 10 (which is superseded and EOL) and CentOS/RHEL 8 (which is superseded but I think still supported), so maybe we don't need to go that far into retrocomputing? The main things we'd lose the ability to compile on by bumping the requirement to 0.55.0 are CentOS/RHEL 7 and Ubuntu 20.04, which have been superseded but I believe are still supported to some extent.

### DIFF
--- a/libportal/meson.build
+++ b/libportal/meson.build
@@ -242,7 +242,7 @@ endif
 # Qt 5 #
 ########
 
-have_cpp = add_languages('cpp', required: get_option('backend-qt5'))
+have_cpp = add_languages('cpp', required: get_option('backend-qt5'), native: false)
 qt5_dep = dependency('qt5', modules: ['Core', 'Gui', 'X11Extras', 'Widgets'], required: get_option('backend-qt5'))
 
 if have_cpp and qt5_dep.found()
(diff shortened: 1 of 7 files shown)


GeorgesStavracas (approved):
LGTM. I think I prefer bumping the Meson requirement to 0.55 instead of supporting older Meson versions.

PR #183, opened 2025-01-14
Add API to register app ID

jadahl:
Meant to be used by host apps to more reliably get an app ID. See https://github.com/flatpak/xdg-desktop-portal/pull/1521.

### DIFF
--- a/libportal/portal.c
+++ b/libportal/portal.c
@@ -386,6 +386,117 @@ xdp_portal_new (void)
   return portal;
 }
 
+static void
+register_cb (GObject      *source_object,
+	     GAsyncResult *result,
+	     gpointer      user_data)
+{
+  g_autoptr(GTask) task = user_data;
+  g_autoptr(GVariant) ret = NULL;
+  g_autoptr(GError) error = NULL;
+
+  ret = g_dbus_connection_call_finish (G_DBUS_CONNECTION (source_object),
+				       result,
+				       &error);
+  if (error)
+    g_task_return_error (task, g_steal_pointer (&error));
+  else
+    g_task_return_boolean (task, TRUE);
+}
+
+/**
+ * xdp_portal_register:
+ * @portal: a [class@Portal]
+ * @app_id: an application ID matching the basename of a desktop file
+ * @cancellable: (nullable): optional [class@Gio.Cancellable]
+ * @callback: (scope async): a callback to call when the request is done
+ * @data: (closure): data to pass to @callback
+ *
+ * Registers the application ID with the org.freedesktop.host.Registry portal.
+ * Doing this when sandboxed is a no-op, but when running on the host, it will
+ * help associating portal requests with an application.
+ *
+ * When the request is done, @callback will be called. You can then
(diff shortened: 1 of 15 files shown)


libportal/portal.c
```diff
@@ -386,6 +386,93 @@ xdp_portal_new (void)
...
+				       result,
+				       &error);
+  if (error)
+    {
+      g_task_return_error (task, g_steal_pointer (&error));
+    }
+  else
+    {
+      g_task_return_boolean (task, TRUE);
+    }
```
GeorgesStavracas, line 408: Nit:

```suggestion
  if (error)
    g_task_return_error (task, g_steal_pointer (&error));
  else
    g_task_return_boolean (task, TRUE);
```

GeorgesStavracas (commented):
LGTM, the only thing I think is missing is the documentation blocks for introspection.

jadahl:
Tweaked the qt examples a bit. The qt5 one now plumbs the app Id via `QApplication`, but couldn't do that in qt6 - I had to register the app ID before doing `QApplication a(argc, argv);` since constructing the `QApplication` already started to talk to the portal (I'm guessing for settings or something).

ilya-fedin:
> The qt5 one now plumbs the app Id via `QApplication`, but couldn't do that in qt6 - I had to register the app ID before doing `QApplication a(argc, argv);` since constructing the `QApplication` already started to talk to the portal (I'm guessing for settings or something).

`desktopFileName` and `setDesktopFileName` are static methods of QGuiApplication (which QApplication inherits from), you can use them before constructing QApplication just fine

smcv:
> `desktopFileName` and `setDesktopFileName` are static methods of QGuiApplication (which QApplication inherits from), you can use them before constructing QApplication just fine

Does this mean it would be more correct to call them as `QGuiApplication::setDesktopFileName(APP_ID)` and `QGuiApplication::desktopFileName()`, rather than `app->setDesktopFileName(APP_ID)` and `app->desktopFileName()`?

ilya-fedin:
> Does this mean it would be more correct to call them as `QGuiApplication::setDesktopFileName(APP_ID)` and `QGuiApplication::desktopFileName()`, rather than `app->setDesktopFileName(APP_ID)` and `app->desktopFileName()`?

Exactly

smcv (changes requested):
Please check whether the Qt code paths can be improved based on what @ilya-fedin said

sithlord48:
`QCoreApplication` is the base class for all Qt Application classes 

The following members are static in `QCoreApplication`

`QCoreApplication::setApplicationName`  /   `QCoreApplication::applicationName` 
`QCoreApplication::setOrganizationDomain` / `QCoreApplication::organizationDomain` 
`QCoreApplication::setOrganizationName` /  `QCoreApplication::organizationName` 

The can be called by any of the Qt Application types set / get the needed info

sithlord48:
Any idea whats up with this PR ? it looks like we will need an appid to use the hotkey portal

Issue #169, opened 2024-09-11
intermittent segfault during automated tests


smcv:
While trying to reproduce #166 interactively, I encountered a segfault in the tests.

To reproduce:

* `mkdir _build`
* `podman run --rm -it -w $(pwd) -v $(pwd):$(pwd):ro -v $(pwd)/_build:$(pwd)/_build:rw debian:sid-slim`
* in the container:
    * `sed -i -e 's/Types:.*/Types: deb deb-src/' /etc/apt/sources.list.d/debian.sources`
    * `apt update`
    * `apt upgrade`
    * `apt build-dep libportal`
    * `meson setup _build`
    * `meson compile -C _build`
    * `meson test -C _build --timeout-multiplier=3 --repeat=20 pytest`

Backtrace:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=11, no_tid=no_tid@entry=0)
    at ./nptl/pthread_kill.c:44
#1  0x00007ff323b3347f in __pthread_kill_internal (signo=11, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007ff323ae4532 in __GI_raise (sig=11) at ../sysdeps/posix/raise.c:26
#3  0x00007ff323ae45d0 in <signal handler called> () at /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007ff3218aaa55 in g_task_get_cancellable (task=0x5f6b636f6d737562) at ../../../gio/gtask.c:1275
#5  0x00007ff3216d0610 in call_returned (object=0x2b32cdd0 [GDBusConnection], result=0x2b2b5fe0, data=0x2b341340)
    at ../libportal/inputcapture.c:284
#6  0x00007ff3218aa393 in g_task_return_now (task=task@entry=0x2b2b5fe0 [GTask]) at ../../../gio/gtask.c:1361
#7  0x00007ff3218ab033 in g_task_return (type=<optimized out>, task=0x2b2b5fe0 [GTask]) at ../../../gio/gtask.c:1430
#8  g_task_return (task=0x2b2b5fe0 [GTask], type=<optimized out>) at ../../../gio/gtask.c:1387
#9  0x00007ff321908ce0 in g_dbus_connection_call_done
    (source=0x2b32cdd0 [GDBusConnection], result=<optimized out>, user_data=0x2b2b5fe0)
    at ../../../gio/gdbusconnection.c:6344
[...]

PR #185, opened 2025-02-10
Assert we only free a Call at an appropriate point in its life cycle

smcv:
* Assert that we only free a Call at an appropriate point in its life cycle
    
    Each Call wraps a single GTask, which is created along with the Call and
    cleared when the Call is destroyed. If the Call has become invalid, that
    would indicate a use-after-free similar to #169, or some other memory
    corruption.

---

On a cloud VM instance that was being used for Debian QA, where (for whatever reason) #169 seems to happen much more often than on the official autobuilders or my own system, this replaces the segfault seen in #169 with an assertion failure: the GTask is no longer a valid object pointer (in fact it happens to be null).

### DIFF
--- a/libportal/account.c
+++ b/libportal/account.c
@@ -36,6 +36,8 @@ typedef struct {
 static void
 account_call_free (AccountCall *call)
 {
+  g_return_if_fail (G_IS_TASK (call->task));
+
   g_debug ("freeing AccountCall");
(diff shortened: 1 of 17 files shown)


smcv:
CI failure is #187, I don't think that's related to what I changed in this PR.

PR #186, opened 2025-02-10
tests: Run pytest with G_DEBUG=fatal-criticals

smcv:
This ensures that we crash out at any sign of undefined behaviour. This is the default for GLib's own GTest framework, but not for other test frameworks like pytest.

Note that we cannot use fatal-warnings (which would be the default for GTest) because at least one test-case for the inputcapture portal emulates a faulty portal implementation, which libportal will log as a g_warning().


smcv:
CI failure is #187, I don't think that's related to what I changed in this PR.

smcv:
Actually it looks as though some of the tests might be triggering criticals under normal circumstances. I think that's a bug (the point of critical warnings is that they're ... critical) but it means we can't apply this just yet.

bbhtt:
Updating the images in CI increased the number of critical warnings..

Issue #190, opened 2025-02-10
inputcapture: Call objects are leaked on successful code paths


smcv:
The memory management of Call objects in the inputcapture interface seems to be wrong in both directions. Not only is there a use-after-free for some orders of events (#169), the Call objects are also frequently leaked.

To reproduce, first apply #189, then apply this:

```diff
diff --git a/libportal/inputcapture.c b/libportal/inputcapture.c
index d452ff2..7f3264c 100644
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -237,6 +237,7 @@ typedef struct {
 
 static void create_session (Call *call);
 static void get_zones (Call *call);
+static size_t call_instances = 0;
 
 static void
 call_free (Call *call)
@@ -264,6 +265,8 @@ call_free (Call *call)
 
   g_free (call->session_path);
 
+  call_instances--;
+  g_debug ("Call instances: %zu", call_instances);
   g_free (call);
 }
 
@@ -277,6 +280,9 @@ call_new (XdpPortal *portal,
 {
   Call *call = g_new0 (Call, 1);
 
+  call_instances++;
+  g_debug ("Call instances: %zu", call_instances);
+
   call->portal = g_object_ref (portal);
 
   if (session != NULL)
diff --git a/tests/meson.build b/tests/meson.build
index d126a06..600fef0 100644
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -17,7 +17,7 @@ if meson.version().version_compare('>= 0.56.0')
 
     test('pytest',
       pytest,
-      args: ['--verbose', '--verbose', '--log-level=DEBUG'],
+      args: ['--verbose', '--verbose', '--log-level=DEBUG', '-s'],
       env: test_env,
       workdir: meson.current_source_dir(),
       timeout: 180,
```

and run the test suite.

Expected result: the number of Call instances "alive" fluctuates between 0 and 1, perhaps sometimes as high as 2 or 3 if GDBus calls are still happening in the background after we have moved on to the next test-case.

Actual result: the number of instances increases over time, only rarely decreasing by 1, and the last count I see in my test log is:

```
(process:386365): libportal-DEBUG: 11:59:31.380: Call instances: 23
```

I think this is because `call_free()` is called when an error occurs (which can result in a double-free like #169 if more than one error is seen), but is not called when we get a successful result and return it from the GTask. If a test-case involves interactions with input capture being completely successful, the Call is just never freed.

[...]

PR #191, opened 2025-02-10
inputcapture: Reference-count Call objects

smcv:
Best reviewed commit-by-commit.

* glib-backports: Add a backport of g_steal_handle_id()
    
    This is useful when writing idempotent code to unsubscribe from GDBus
    signals.

* inputcapture: Factor out call_dispose() from call_free()
    
    This makes call_dispose() idempotent, so that it can be called whenever a
    Call has become irrelevant, but without freeing the memory used for the
    Call structure itself, which would leave it as a dangling pointer if it
    is still in use as the user-data of an async call.
    
    This will be used in a subsequent commit to keep the Call alive via a
    circular reference between the GTask and Call while we are still waiting
    for a result, and then break the circular reference with call_dispose()
    when the Call has reached a resolution, either success or failure.

* inputcapture: Reference a pre-existing session where necessary
    
    We're passing the session object to g_task_return_pointer() with
    g_object_unref as the free-function, and that's only correct if we
    previously owned a reference to it. In the case where we created a fresh
    session object, we did own a reference, but in the case where we merely
    retrieved it from call->session, we did not.
    
    This would have led to a use-after-free if it wasn't for the fact that
    we are leaking the Call object (#190), which leaks a reference to the
    session, avoiding the use-after-free via compensating errors. Before we
    can safely fix the leak, we have to reference-count the session correctly.

* inputcapture: Reference-count Call instances
    
    To avoid a use-after-free (#169), we need to make sure that there is
    at least one reference held to the memory used to store the Call struct
    for the duration of each D-Bus method call, so consistently take a new
    ref to the user-data of g_dbus_connection_call(), and release it in
    call_returned().
    
    However, we also need to keep the Call alive for as long as it is
    listening to any D-Bus signals, without introducing a long-term
    circular reference that would make it be leaked indefinitely.
    
    To achieve this, create a temporary circular reference between the Call
    and the GTask, but break it via call_dispose() every time we reach a
    resolution to the task, whether that's success or a failure.
    
    Resolves: https://github.com/flatpak/libportal/issues/169
    Resolves: https://github.com/flatpak/libportal/issues/190

* inputcapture: Assert that we don't prep_call() twice in parallel
    
    If we did, then the second signal subscription would overwrite the first,
    and call_dispose() would no longer unsubscribe the first signal
    subscription, breaking our memory-management assumptions.

* inputcapture: Drop unused userdata parameter to prep_call
    
    Despite its name, this was really a free-function for the GDBus
    signal subscription, which in practice was always NULL.

* inputcapture: Guard against unknown Response codes
    
    If the response code was somehow neither 0, 1 nor 2, we would previously
    have freed the GTask without ever returning a result. Return a failed
    result in this case, mechanically equivalent to response code 2
    ("The user interaction was ended in some other way") but with a
    different error message.

* inputcapture: Consistently end the Call iff the GTask has completed
    
    After the previous commit ensured that the GTask is always given a
    result at all appropriate times, this should not make any practical
    difference.
    
    However, it hopefully makes the intention clearer: after the GTask has
    completed, we always dispose the Call, and if the GTask has not yet
    completed, we never do.

* inputcapture: Assert that we don't receive Response signals after dispose
    
    When the Call object is disposed, it unsubscribes from the Response
    signal before releasing its GTask reference, so it should be impossible
    to get into the Response handler after dispose. Clarify that.

### DIFF
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -239,8 +239,10 @@ static void create_session (Call *call);
 static void get_zones (Call *call);
 
 static void
-call_dispose (Call *call)
+call_dispose (void *p)
 {
+  Call *call = p;
+
   /* CreateSesssion */
   if (call->parent)
     call->parent->parent_unexport (call->parent);
@@ -271,18 +273,9 @@ call_ref (Call *call)
 }
 
 static inline void
-call_last_unref (void *call)
-{
-  /* Only separated from call_dispose() because g_rc_box_release_full()
-   * wants a GDestroyNotify, and because this is a convenient place to put
-   * life-cycle debugging */
-  call_dispose (call);
-}
-
-static inline void
 call_unref (void *call)
 {
-  g_rc_box_release_full (call, call_last_unref);
+  g_rc_box_release_full (call, call_dispose);
 }
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (Call, call_unref)
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -544,6 +544,10 @@ get_zones_done (GDBusConnection *bus,
   guint32 response;
   g_autoptr(GVariant) ret = NULL;
 
+  /* If the Call has already been disposed, we should have unsubscribed
+   * from the Response signal at that time, so we shouldn't get here */
+  g_return_if_fail (G_IS_TASK (call->task));
+
   g_variant_get (parameters, "(u@a{sv})", &response, &ret);
 
   if (response != 0)
@@ -676,6 +680,10 @@ session_created (GDBusConnection *bus,
   Call *call = data;
   guint32 response;
   g_autoptr(GVariant) ret = NULL;
+
+  /* If the Call has already been disposed, we should have unsubscribed
+   * from the Response signal at that time, so we shouldn't get here */
+  g_return_if_fail (G_IS_TASK (call->task));
 
   g_variant_get (parameters, "(u@a{sv})", &response, &ret);
 
@@ -955,6 +963,10 @@ set_pointer_barriers_done (GDBusConnection *bus,
   g_autoptr(GVariant) ret = NULL;
   GVariant *failed = NULL;
   GList *failed_list = NULL;
+
+  /* If the Call has already been disposed, we should have unsubscribed
+   * from the Response signal at that time, so we shouldn't get here */
+  g_return_if_fail (G_IS_TASK (call->task));
 
   g_variant_get (parameters, "(u@a{sv})", &response, &ret);
 
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -633,6 +633,8 @@ get_zones_done (GDBusConnection *bus,
     g_task_return_new_error (call->task, G_IO_ERROR, G_IO_ERROR_CANCELLED, "InputCapture GetZones() canceled");
   else if (response == 2)
     g_task_return_new_error (call->task, G_IO_ERROR, G_IO_ERROR_FAILED, "InputCapture GetZones() failed");
+  else if (response != 0)
+    g_task_return_new_error (call->task, G_IO_ERROR, G_IO_ERROR_FAILED, "InputCapture GetZones() unknown response code %d", response);
 
   /* If the Call failed, we can ignore any subsequent method replies */
   if (response != 0)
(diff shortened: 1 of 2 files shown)


smcv:
Marked as draft because I'm still testing this (100 runs in a loop) on a cloud VM instance that, for whatever reason, reproduces #169 more easily than any other environment I have access to. But I think it's ready for review.

It certainly does resolve #191. When I add this:

```diff
diff --git a/libportal/inputcapture.c b/libportal/inputcapture.c
index 0736a0e..1504cf1 100644
--- a/libportal/inputcapture.c
+++ b/libportal/inputcapture.c
@@ -237,6 +237,7 @@ typedef struct {
 
 static void create_session (Call *call);
 static void get_zones (Call *call);
+static size_t call_instances = 0;
 
 static void
 call_dispose (Call *call)
@@ -277,6 +278,8 @@ call_last_unref (void *call)
    * wants a GDestroyNotify, and because this is a convenient place to put
    * life-cycle debugging */
   call_dispose (call);
+  call_instances--;
+  g_debug ("Call instances: %zu", call_instances);
 }
 
 static inline void
@@ -297,6 +300,9 @@ call_new (XdpPortal *portal,
 {
   g_autoptr(Call) call = g_rc_box_new0 (Call);
 
+  call_instances++;
+  g_debug ("Call instances: %zu", call_instances);
+
   call->portal = g_object_ref (portal);
 
   if (session != NULL)
diff --git a/tests/meson.build b/tests/meson.build
index d126a06..600fef0 100644
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -17,7 +17,7 @@ if meson.version().version_compare('>= 0.56.0')
 
     test('pytest',
       pytest,
-      args: ['--verbose', '--verbose', '--log-level=DEBUG'],
+      args: ['--verbose', '--verbose', '--log-level=DEBUG', '-s'],
       env: test_env,
       workdir: meson.current_source_dir(),
       timeout: 180,
```

I can still see the number of instances increase, but only gradually (the highest it reaches is 6), and the number drops back down to 0 by the time testing finishes.

smcv:
cc @whot

smcv:
> Marked as draft because I'm still testing this (100 runs in a loop) on a cloud VM instance that, for whatever reason, reproduces https://github.com/flatpak/libportal/issues/169 more easily than any other environment I have access to

It was successful. I'm continuing to test in the same environment with the whole test suite.

smcv:
Rebased after @TingPing merged prerequisites #184 and #189 (thanks!)

A previous version survived 100 iterations of repeated testing on a cloud VM where 0.9.0 hits #169 80% of the time. Testing the rebased version now, it will likely take a while but is looking good so far (on its 10th run).

smcv:
> Testing the rebased version now, it will likely take a while but is looking good so far (on its 10th run)

This also survived 100 test iterations, on a cloud VM where 0.9.0 had approximately an 80% failure rate, so I'm now quite confident that this PR fixes the race condition.

libportal/inputcapture.c
```diff
@@ -264,13 +264,29 @@ call_dispose (Call *call)
...
+   * wants a GDestroyNotify, and because this is a convenient place to put
+   * life-cycle debugging */
   call_dispose (call);
-  g_free (call);
 }
 
+static inline void
+call_unref (void *call)
+{
+  g_rc_box_release_full (call, call_last_unref);
```
TingPing, line 285: This is nitpicking, but `g_rc_box_release_full` refers to the destroy notify as a *"clear"* function, which IME is the convention.

On its own that doesn't matter but your usage of `call_dispose()` below would be more readable to me if those were `call_clear()`.

TingPing (approved):
>  inputcapture: Consistently end the Call iff the GTask has completed

Typo: "iff"

Everything looks good to me.

smcv:
> Typo: "iff"

No, that was intentional: it's mathematical jargon for "if and only if", which would have made the subject line of the commit too long.

If you don't like that, perhaps `inputcapture: Consistently end Call if and only if GTask completed` would be better?

libportal/inputcapture.c
smcv: I did originally call the function that clears all the fields `call_clear()`, but later renamed it to `call_dispose()`. My reasoning for that was that "clear" functions are typically only used when an object is about to be destroyed, but in this case a `Call` can hang around for a significant timespan (up to 25 seconds if the D-Bus call times out) after it has already been cleared, and we need to make sure not to crash.

I chose `call_dispose()` to suggest a similarity to GObject `dispose`, and its use in breaking reference cycles by calling `g_object_run_dispose()` (which is the equivalent of what we're doing every time we call it *before* last-unref).

But if you still think `call_clear()` would be clearer (no pun intended) after seeing that reasoning, I can rename it.

The separation between `call_last_unref()` and `call_dispose()` was mainly so that I could add debug instrumentation in `call_last_unref()` (https://github.com/flatpak/libportal/pull/191#issuecomment-2647911668), which I used to verify that #190 is fixed by this branch; but now that I've done that verification and I'm only making cosmetic changes, I can combine them into one function named `call_dispose()` or `call_clear()` if you'd prefer.

TingPing: I'm fine with `call_dispose` given your reasoning.

If its not a burden probably should remove `call_last_unref`.

whot (approved):
Thanks heaps for finding all these... quite a ride just reading the commit messages :)

PR #202, opened 2025-06-12
remote: update internal state after condition checks

andyholmes:


bbhtt:
It would be nice to have some description in the commit body what this is fixing.

PR #207, opened 2025-09-30
build-aux: Use GNOME 49 runtime for Gtk 3 example, too

smcv:
---

@TheEvilSkeleton, was there a reason you didn't do this already in #206?

### DIFF
--- a/build-aux/org.gnome.PortalTest.Gtk3.json
+++ b/build-aux/org.gnome.PortalTest.Gtk3.json
@@ -1,7 +1,7 @@
 {
     "app-id": "org.gnome.PortalTest.Gtk3",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "47",
+    "runtime-version": "49",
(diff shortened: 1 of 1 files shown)


TheEvilSkeleton:
Yes, because the window explicitly states "NEW VERSION":

https://github.com/flatpak/libportal/blob/8cfddc06951162c24d9e9ddee262fe55c65fdd65/portal-test/gtk3/portal-test-win.ui#L98

Since it wants us to use the GTK4 frontend, I didn't bother bumping the runtime for the GTK3 version.

smcv:
That's been there since several years before the Gtk4 example was added. I don't think it means "you should use the new version instead", I think it means something more like "this is the new version of the Gtk3 portal test (integrated into libportal), not the old version (a standalone app)".

libportal should continue to have a Gtk 3 UI for as long as it has `portal-gtk3.h`, so that there's an easy way to do manual testing.

PR #212, opened 2026-01-09
Update minimum required Meson version to 1.9.0

nda-cunh:
image

references:
https://mesonbuild.com/Release-notes-for-1-9-0.html#pkgconfiggenerate-supports-internal-dependencies-in-requires

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('libportal','c',
         version: '0.9.2',
         default_options: ['warning_level=2'],
-        meson_version: '>= 0.55.0')
+        meson_version: '>= 1.9.0')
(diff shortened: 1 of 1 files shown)


nda-cunh:
Oh, sorry, I closed the merge request on WrapDB by mistake. In any case, this issue isn’t really related to WrapDB or Meson; it’s mainly that the project has specified the wrong version of meson.build, as the Meson used by LibPortal utilises features from Meson 1.9.0, whilst it states that it requires Meson 0.55, which is incorrect. So either we need to change the features to revert to Meson 0.55, or we need to request a more recent version of Meson!

PR #208, opened 2025-11-20
Add clipboard API and new input capture session construction API

jadahl:
This adds clipboard API usable with the existing remote desktop portal, and implements the new split session construction input capture API, as well as input capture clipboard support. See https://github.com/flatpak/xdg-desktop-portal/pull/1803.

PR #214, opened 2026-01-30
InputCapture support for session persistence

whot:
This is on top of @jahdal's work from https://github.com/flatpak/libportal/pull/208

See https://github.com/flatpak/xdg-desktop-portal/pull/1898 for references to other components.


KoljaFrahm:
Seems like the fix is merged, thank you!

sithlord48:
Im looking forward to this if only to stop the almost daily AI generated PR's in deskflow to add this.

whot:
There's a little tool to test all this a bit easier now: https://github.com/whot/libportal-clipboard-test

And with #217 fixed valgrind is happy now, so this should be good to go.

libportal/inputcapture.c
```diff
@@ -963,6 +970,14 @@ start_session (Call *call)
   g_variant_builder_add (&options, "{sv}", "capabilities",
                          g_variant_new_uint32 (call->capabilities));
 
+  if (call->session->persistence != XDP_INPUT_CAPTURE_SESSION_PERSISTENCE_NONE)
+    g_variant_builder_add (&options, "{sv}", "persist_mode",
+                           g_variant_new_uint32 (call->session->persistence));
+
+  if (call->session->restore_token)
+    g_variant_builder_add (&options, "{sv}", "restore_token",
+                           g_variant_new_string (call->session->restore_token));
```
jadahl, line 1072: Should this do a version check on the portal first? The clipboard one doesn't, but it also is just a request method that may fail, so it's harmless to do without checking the version. Here one would have to avoid setting these for older versions since otherwise the request I assume would fail.

sithlord48:
Should the methods names here be considered  "final" ?

jadahl:
> Should the methods names here be considered "final" ?

You mean they can't be renamed? I'd consider them final once the pull request lands and there is a release; before that things not yet added can be renamed.

sithlord48:
> > Should the methods names here be considered "final" ?
> 
> You mean they can't be renamed? I'd consider them final once the pull request lands and there is a release; before that things not yet added can be renamed.

After this lands is there a plan for release ?

jadahl:
> After this lands is there a plan for release ?

That be a sensible thing to do at least.

libportal/inputcapture.c
whot: done (as follow-up to make it easier to review), and I added a bunch of test cases too

jadahl (approved):

sithlord48:
LGTM

KoljaFrahm:
The PR is tested and approved, so can you merge it?
And please do a release after

skewty:
Tagging @smcv (Simon) who was instrumental in last version release. Are we good to go for another release? If not, what is missing / required?

mikeeq:
@smcv @GeorgesStavracas @jadahl let us know if you need any help with releasing a new version, i.e.: if there's anything outstanding which you planned to address before proceeding with a release.

jadahl:
I think it'd make sense to create a libportal release roughly the same time as the next xdg-desktop-portal release is made that contains the new interfaces that this adds APIs for.

PR #220, opened 2026-05-18
portal: Fix a refcounting issue

whot:
We add the session to our hash table but also pass it back to the caller. If the caller unrefs the session we're left with a dangling pointer which will eventually cause issues - latest when we g_object_unref() it during our own cleanup.

Fixes: 811d636d2c99 ("portal: Keep track of sessions")

cc @jadahl


jadahl (approved):

bbhtt:
I assume this is desired before a release.

PR #221, opened 2026-05-31
gir: Specify --doc-format=gi-docgen

A6GibKm:
See
https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/401.

Note that version_compare() requires meson 0.62.

### DIFF
--- a/libportal/meson.build
+++ b/libportal/meson.build
@@ -98,6 +98,12 @@ libportal_dep = declare_dependency(
 )
 
 if introspection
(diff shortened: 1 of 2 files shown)


A6GibKm:
Unfortunately, the meson version in Ubuntu 22.04 is too old.

KoljaFrahm:
Now there is no more Ubuntu 22.04 CI

A6GibKm:
CI passes now.

PR #233, opened 2026-09-08
Release 0.11.0

bbhtt:


meson.build
```diff
@@ -1,5 +1,5 @@
 project('libportal','c',
-        version: '0.10.1',
+        version: '0.11.1',
```
GeorgesStavracas, line 2: 0.11.0

bbhtt: It's in the first commit, I tag it from the first commit.

GeorgesStavracas: Ah sorry, heh GitHub combined the diffs and only showed the final output.

GeorgesStavracas (approved):

KoljaFrahm (approved):

jadahl:
Seems the pipeline (https://github.com/flatpak/libportal/actions/runs/34251069735) is stuck.

sithlord48:
@bbhtt  Can you return the failed job, it looks to be just a random upload error that caused the failure

